PR #1, opened 2013-07-18
MPLUGIN-248: Fix all ITs to use the correct XML namespace.

mfriedenhagen:
See http://jira.codehaus.org/browse/MPLUGIN-248


michael-o:
Fixed in 3.3.

jira-importer:
Resolve #875

PR #2, opened 2013-07-30
MPLUGIN-244  Help mojo generates Javadoc, which is not accepted by JDK 8...

muminc:
The <br/> close tag is causing Java 8 doclint to fail.

Stackoverflow http://stackoverflow.com/questions/13720626/is-javadoc-html-or-xhtml article suggests JavaDoc must be in html as opposed to xhtml.


michael-o:
Already fixed in 3.3

jira-importer:
Resolve #859

PR #5, opened 2015-09-22
Preserving the annotations at runtime is very useful for testing 

stickycode:
I want to run a mojo in a harness for testing that can manage a variety of configurations, but relies on the annotation metadata to identify configuration points. 

I can't think of a good reason for not having these preserved at runtime. Hopefully this submission is a no brainer.

### DIFF
--- a/maven-plugin-annotations/pom.xml
+++ b/maven-plugin-annotations/pom.xml
@@ -37,5 +37,11 @@
       <artifactId>maven-artifact</artifactId>
       <version>3.0</version>
     </dependency>
+    <dependency>
+      <groupId>junit</groupId>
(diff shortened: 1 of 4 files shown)


ifedorenko:
You can use ASM or similar tool to inspect annotation present in .class files.

stickycode:
I could but these annotations make sense to be preserved at runtime and read via reflection. Its simple and obvious.

ifedorenko:
sure, I am not against retaining annotations at runtime, just saying you already can write the tests you want.

michael-o:
@stickycode Please create a JIRA issue for this to be merged.

PR #10, opened 2018-01-02
[MPLUGIN-331] Let plugin:report support takari-maven-plugin packaging

ppalaga:


maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/PluginReport.java
```diff
@@ -197,9 +197,18 @@
     @Component
     private RuntimeInformation rtInfo;
 
+    /**
+     * Path to {@code plugin.xml} plugin descriptor to generate the report from.
+     *
+     * @since 3.5.1
+     */
+    @Parameter( defaultValue = "${project.build.outputDirectory}/META-INF/maven/plugin.xml", required = true )
```
rfscholte: I would like to see this as a readOnly parameter. Changing its value will break things.

ppalaga:
> Changing its value will break things.

Which things will it break?

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/PluginReport.java
ppalaga: Which things will it break?

rfscholte: https://maven.apache.org/ref/3.5.2/maven-plugin-api/ says that the `plugin.xml` is alwys expected at this location. So it should be considered a contant, not a variable, hence why I would like to see it as readOnly.

ppalaga: Thanks for the explanation, @rfscholte . But https://maven.apache.org/ref/3.5.2/maven-plugin-api/ seems to define the the location of the `plugin.xml` within the `jar` rather than its location in the source tree. This `pluginXmlFile` indeed is a location in the source tree and this `PluginReport` is a report mojo that will not impact the location of `plugin.xml` in the jar in any way. That said, I cannot see how a writable `pluginXmlFile` can violate the maven-plugin-api.

rfscholte: The claim of breaking things is a bit too hard, but making it variable assumes that is configurable in real world. It will also appear on reports like the goal page, the help goal, etc.
Let's not confuse people or show parameters which will only distract them.
What is the issue of *not* making it readOnly?

hboutemy:
Peter: IIUC, making this location configurable is not useful to your use case, isn't it?
Then it's better to not add an additional parameter that nobody will ever use: it's just noise and complexity

Without objection, I'll do the job in a few days (and do a release :) )

ppalaga:
@hboutemy @rfscholte sorry for the late response. I have to admit, I have envisaged the configurable `pluginXmlFile` to eventually be able to use an alternative `plugin.xml` only to produce the site during a release and keep using takari-lifecycle-plugin for jar. The takari-lifecycle-plugin produces an imperfect `plugin.xml` (see [1]) and the shortages matter only during the site production. This being only an initial idea rather than a proven solution how to fix my situation, I agree to either make the param readonly or remove the param altogether and hard code the plugin.xml path instead. Thanks for your patience.

[1] https://github.com/takari/takari-lifecycle/issues/63

hboutemy:
ok, I'll mark it as read only

jira-importer:
Resolve #640

PR #11, opened 2018-01-03
MPLUGIN-328 ArrayIndexOutOfBoundsException: 48188

dheid:
If a dependency contains a class that could not be read by the default mojo annotation scanner, Maven will fail. Especially classes compiled using AspectJ are affected. Please consider merging this pull request, as it just lets Maven print a warning and won't let it interrupt the build process.

Fixes https://issues.apache.org/jira/browse/MPLUGIN-328


dheid:
Maintainer fixed it on his own.

jira-importer:
Resolve #636

PR #12, opened 2018-02-16
Add @Generated annotation to Mojo

gaul:
This allows tools like error-prone to ignore the generated code.

### DIFF
--- a/maven-plugin-tools-generators/src/main/resources/help-class-source.vm
+++ b/maven-plugin-tools-generators/src/main/resources/help-class-source.vm
@@ -32,6 +32,7 @@ import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
 import org.xml.sax.SAXException;
 
+import javax.annotation.Generated;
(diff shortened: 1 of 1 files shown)


Tunaki:
Unfortunately, I'm not sure if this solves more problems than it creates... There are lots of issues with this annotation and Java 9 for tools that read Java source code. I just tried adding it and then generating the Javadoc of a plugin, it failed with:

```
[ERROR] import javax.annotation.Generated;
[ERROR]             ^
[ERROR]   (package javax.annotation is declared in module java.xml.ws.annotation, which is not in the module graph)
[ERROR]
[ERROR] Command line was: "C:\Program Files\Java\jdk-9.0.1\bin\javadoc.exe" @options @packages
```

There is also a new [`javax.annotation.processing.Generated`](https://docs.oracle.com/javase/9/docs/api/javax/annotation/processing/Generated.html) annotation since Java 9...

In any case, if something should be done here, modular compatibility is priority.

elharo:
Please open  JIRA issue for this if it's still important.

PR #14, opened 2018-07-16
[MPLUGIN-336] Upgrade ASM to 6.2 for Java 10,11 support

zregvart:
This upgrades ASM to 6.2 for Java 10 and early Java 11 support.

### DIFF
--- a/maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/PluginHelpGenerator.java
+++ b/maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/PluginHelpGenerator.java
@@ -36,8 +36,8 @@ import org.codehaus.plexus.velocity.VelocityComponent;
 import org.objectweb.asm.ClassReader;
 import org.objectweb.asm.ClassVisitor;
 import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.commons.ClassRemapper;
 import org.objectweb.asm.commons.Remapper;
-import org.objectweb.asm.commons.RemappingClassAdapter;
 import org.objectweb.asm.commons.SimpleRemapper;
 
 import java.io.File;
@@ -421,7 +421,7 @@ public class PluginHelpGenerator
 
         Remapper packageRemapper =
             new SimpleRemapper( HELP_MOJO_CLASS_NAME, packageAsDirectory + '/' + HELP_MOJO_CLASS_NAME );
-        ClassVisitor cv = new RemappingClassAdapter( cw, packageRemapper );
+        ClassVisitor cv = new ClassRemapper( cw, packageRemapper );
 
         try
         {
--- a/pom.xml
+++ b/pom.xml
@@ -97,7 +97,7 @@
     <antVersion>1.7.1</antVersion>
     <mavenInvokerPluginVersion>3.0.1</mavenInvokerPluginVersion>
     <maven.site.path>plugin-tools-archives/plugin-tools-LATEST</maven.site.path>
-    <asmVersion>5.0.2</asmVersion>
+    <asmVersion>6.2</asmVersion>
   </properties>
 
   <dependencyManagement>
@@ -291,6 +291,21 @@
     <plugins>
       <plugin>
         <artifactId>maven-enforcer-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>enforce-bytecode-version</id>
+            <configuration>
+              <rules>
+                <enforceBytecodeVersion>
+                  <excludes>
+                    <!-- ASM supports Java 1.5+, see https://asm.ow2.io/faq.html#Q14 -->
+                    <exclude>org.ow2.asm:*</exclude>
+                  </excludes>
+                </enforceBytecodeVersion>
+              </rules>
+            </configuration>
+          </execution>
+        </executions>
         <dependencies>
           <dependency>
             <groupId>org.codehaus.mojo</groupId>
@@ -397,42 +412,6 @@
           </plugin>
         </plugins>
       </reporting>
-    </profile>
-    <profile>
-      <activation>
-        <jdk>[1.9,)</jdk>
-      </activation>
-      <build>
-        <plugins>
-          <plugin>
-            <artifactId>maven-enforcer-plugin</artifactId>
-            <executions>
-              <execution>
-                <id>enforce-bytecode-version</id>
-                <configuration>
-                  <rules>
-                    <enforceBytecodeVersion>
-                      <ignoreClasses>
-                        <ignoreClass>module-info</ignoreClass>
-                      </ignoreClasses>
-                    </enforceBytecodeVersion>
-                  </rules>
-                </configuration>
-              </execution>
-            </executions>
-            <dependencies>
-              <dependency>
-                <groupId>org.codehaus.mojo</groupId>
-                <artifactId>extra-enforcer-rules</artifactId>
-                <version>1.0-beta-5</version>
-              </dependency>
-            </dependencies>
-          </plugin>
-        </plugins>
-      </build>
-      <properties>
-        <asmVersion>6.1.1</asmVersion>
-      </properties>
     </profile>
   </profiles>
 </project>


Godin:
FWIW just dependency upgrade might be not enough for JDK 11 - [`ClassVisitor` without `Opcodes.ASM7_EXPERIMENTAL` throws `UnsupportedOperationException`](https://gitlab.ow2.org/asm/asm/blob/ASM_6_2/asm/src/main/java/org/objectweb/asm/ClassVisitor.java#L157) when [`ClassReader` encounters/visits nest mates](https://gitlab.ow2.org/asm/asm/blob/ASM_6_2/asm/src/main/java/org/objectweb/asm/ClassReader.java#L541)

zregvart:
@Godin The error I got on Apache Camel did not occur with this change. We target Java 8 classes and wanted to run tests on Java 11. The issue might occur only if you target Java 10+ class versions.
I think this should target the use cases without [JEP-181](http://openjdk.java.net/jeps/181) until ASM API for nests stabilises.

sormuras:
ASM 6.2.1 was released.

Tried to compile a plugin with JDK 11 and `--release 11` yielded:

```text
...
Caused by: java.lang.UnsupportedOperationException
    at org.objectweb.asm.ClassVisitor.visitNestHostExperimental (ClassVisitor.java:158)
    at org.objectweb.asm.ClassReader.accept (ClassReader.java:541)
    at org.objectweb.asm.ClassReader.accept (ClassReader.java:391)
    at org.apache.maven.tools.plugin.extractor.annotations.scanner.DefaultMojoAnnotationsScanner.analyzeClassStream (DefaultMojoAnnotationsScanner.java:219)
    at org.apache.maven.tools.plugin.extractor.annotations.scanner.DefaultMojoAnnotationsScanner.scanDirectory (DefaultMojoAnnotationsScanner.java:200)
    at org.apache.maven.tools.plugin.extractor.annotations.scanner.DefaultMojoAnnotationsScanner.scan (DefaultMojoAnnotationsScanner.java:108)
    at org.apache.maven.tools.plugin.extractor.annotations.scanner.DefaultMojoAnnotationsScanner.scan (DefaultMojoAnnotationsScanner.java:84)
    at org.apache.maven.tools.plugin.extractor.annotations.JavaAnnotationsMojoDescriptorExtractor.scanAnnotations (JavaAnnotationsMojoDescriptorExtractor.java:124)
    at org.apache.maven.tools.plugin.extractor.annotations.JavaAnnotationsMojoDescriptorExtractor.execute (JavaAnnotationsMojoDescriptorExtractor.java:103)
    at org.apache.maven.tools.plugin.scanner.DefaultMojoScanner.populatePluginDescriptor (DefaultMojoScanner.java:96)
...
```

sormuras:
> ... until ASM API for nests stabilises.

@forax Is ASM 6.2.1 ready for nest support? Or will there be another bug-fix/improvement release addressing nest issues?

forax:
ASM 6.2.1 (and ASM 6.2) are nestmate ready, but you have to opt-in using the api ASM7_EXPERIMENTAL instead of the api ASM6 (cf @Godin answer).
 
To follow the JDK cadence release, we now introducing features even if the corresponding JDK have not yet released (the spec part has to be finished obviously) but under an EXPERIMENTAL flag, so for ASM 6.2 we have introduce the support of all 3 Java 11 features if you use ASM7_EXPERIMENTAL.
ASM 6.2.1 adds bugfixes and the support of Java 12 classfile version.  

After the first release candidate of the JDK, so later this month or at the beginning of September, we will release ASM7, remove the support of ASM7 and add the support of ASM8_EXPERIMENTAL as soon as a feature of Java 13 that impact the classfile will be fully specified (it's Java 13 because currently no VM spec change are scheduled for Java 12).

sormuras:
Thanks for the prompt and detailed reply.

docrjp:
So how is the visitNestHostExperimental() call coming about, if it's an experimental feature that needs opting in? The code in Maven Plugin Tools is this:

```java
ClassReader rdr = new ClassReader( is );
rdr.accept( mojoClassVisitor, ClassReader.SKIP_FRAMES | ClassReader.SKIP_CODE | ClassReader.SKIP_DEBUG );
```

And the Mojo visitors don't appear to be doing anything to cause visitNestHostExperimental() to be called either. And there is no call to this method or any reference to nests in maven plugin tools.

So it would seem ASM is visiting the nests, when ASM7_EXPERIMENTAL isn't enabled, and choking on it's own decision? More like an IllegalStateException?

forax:
The ClassReader reads a class attribute named NestHost because javac 11 now emits such attribute, but because the mojoClassVisitor is configured with ASM6 (at construction) it fails because ASM knows that the visitor doesn't handle such attribute.

ASM doesn't let a visitor ignore nest mate attributes because those attributes enable/disable access to private members from an inner class, so if by default the visitor ignore this attribute you will get weird IllegalAccessError later at runtime.

So the implementer of the mojo visitor has to decide if it's ok to just enabling ASM7_EXPERIMENTAL and if the methods visitNestHostExperimental() and visitNestMembersExperimental have to be implemented.

docrjp:
@forax that makes a lot of sense, thanks.

I can see that Jacoco and cglib adopted ASM7_EXPERIMENTAL, which bodes well that the approach ASM is taking will get picked up. If the official Maven plugins also did this, then I would say that a very large class of downstream projects would then open up to development against early access releases of the JDK. That would be a very positive development all round for Java given the new release cadence and Oracle's changing approach to LTS. I think otherwise, we'll be in a pretty unsettling situation where a new Java version drops, the old one is considered out-of-date, but Maven projects would only just begin the process of updating (besides perhaps targeting an old version and forking tests on the new runtime). And this will happen every 6 months potentially.

How realistic would it be for Maven to provide experimental support in kind?

rfscholte:
The Java Release Train comes with a lot of challenges, whereas staying in line with ASM is a clear example. I understand there are projects that want to experiment with new ASM features, but that might imply extra codechanges. (a request for having an option to get the latest ASP api version was declined)

Is it important to have a released version of the plugin? Otherwise you can fork the plugin yourself and implement this yourself.

Another option I'm thinking of is to introduce an interface, so required changes from ASM can be implemented by anybody, then it will be a matter of adding this dependency to the plugin in your own pom.

maven-plugin-plugin is just one of the plugins, there are more plugins suffering with the same issue and I don't expect that we will do releases on demand, our team is just too small and the project too big.

I'm open for any solution that won't require a new release of this plugin with every change (new API) of ASM

docrjp:
@rfscholte indeed but it's not about experimenting with ASM or doing releases on demand. Rather it is Maven blocking projects from using Java release candidates.

Most Maven users won't be working with ASM directly, and different Maven plugins as well as runtime can use different ASM versions anyway, so I don't think there are really any issues nor expectations there.

It's also probably not that big a deal. I would think that any read-only use of ASM (i.e. ClassReader but not ClassWriter) would only need to enable the appropriate experimental opcode, and we're done. Certainly, that is all Jacoco and Cglib had to do. In fact, cglib provides a way to configure it:

https://github.com/cglib/cglib/commit/fd9f5021c94250baf8f2701576cc8db1a500908e

I think something of this kind would solve most use cases, and I'm happy to help bring that about. But, we would want to design a solution first that is sustainable, so I can't just submit some patch requests. We need to collectively come to a solution. If we made it configurable, how would that look? Could ASM provide a stable ASM_EXPERIMENTAL constant so that Maven doesn't have to change every time ASM updates the constant? This isn't about JDK 11 but every Java version going forward and how ASM and Maven prepare for them. That's got to be worth addressing with a solid solution which we all buy into.

rfscholte:
@forax told me I should not solve this as done by cglib, see https://gitlab.ow2.org/asm/asm/issues/317849

forax:
A better solution IMO is to move all ASM related codes (codes that deals with classfiles) into one new module so you only have to update that module once for the benefit of all other maven modules.

docrjp:
That would be good!

But are the different use cases for ASM in Maven sufficiently similar to do so? Probably stating the obvious, but creating any ASM-like visitor pattern on top of ASM is merely moving the problem on; and all the analysis that is quite correct of ASM would apply to the Maven shared code also; i.e. we wouldn't want hard-to-debug errors due to silently ignoring incompatible changes. So are we talking about literally moving all the code and then those ASM-related aspects of the affected maven plugins are done in a shared library, even if they are very specific use cases?

I suppose that would be OK. The uses of ASM in Maven plugin tools, would be annotation analysis and use of an ASM remapper. In plexus-languages, ASM is used to read module-info.java. These are already quite specific use cases in shared utility projects.

How do we enumerate all other cases for official maven plugins?

forax:
@docrjp,
yes the idea is just to consolidate all use cases inside one module,
to enumerate all maven plugins that are using ASM, you can take a look to their dependencies :)

rfscholte:
I've been able to reduce the need for ASM in plexus-java. First implementations always used ASM to read the module descriptor, now that's only done if the runtime is Java 8 or below, otherwise it'll use the `ModuleDescriptor.read()` from the JRE. 
Plugins that still need ASM ar at least maven-plugin-plugin, maven-shade-plugin, maven-dependency-plugin and probably a few more.

forax:
@rfscholte,
there is a slight difference between using ASM or ModuleDescriptor.read is that ModuleDescriptor.read only accept identifiers that are Java valid identifiers, not any strings like ASM does.

rfscholte:
@forax I understand and that's fine. In the end it'll be the JRE complaining with its own clear messages if the result of plexus-java's LocationManager is wrong due to invalid identifiers.

docrjp:
That's great! I will look through maven modules in the following artifact groups and try to document each use of ASM:

org.apache.maven.plugins
org.codehaus.plexus

Let me know if there any other artifact groups we should look at...

sormuras:
Any news regarding this issue? GA of _"openjdk 11-ea 2018-09-25"_ is only two weeks ahead. ;-)

frankgrimes97:
OpenJDK 11 has now been released... any ETA on when we might expect this to be addressed?

rfscholte:
@forax can we expect a new version of ASM that is finalized for Java 11 or is 6.2.1 the preferred dependency?

forax:
The new version of ASM, ASM7 is scheduled for the next week-end.

forax:
hum, correction,
we have done some big chance since ASM 6.2, so we are releasing ASM7 BETA now and will go to a final version in a few weeks.

olamy:
I'd like to get rid of this pr in favour of #15 (more simple and remove the legacy javadoclet)

sormuras:
I agree with @olamy

olamy:
replaced by #15

jira-importer:
Resolve #645

PR #15, opened 2018-10-03
[MPLUGIN-336] Jdk11 asm7

olamy:


olamy:
@rfscholte  @khmarbaise @hboutemy @sormuras  if you can review that (I even try to follow checkstyle :P )

maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/PluginXdocGenerator.java
```diff
@@ -405,8 +395,7 @@ private void writeGoalAttributes( MojoDescriptor mojoDescriptor, XMLWriter w )
      */
     private void writeGoalParameterTable( MojoDescriptor mojoDescriptor, XMLWriter w )
     {
-        @SuppressWarnings( "unchecked" )
-        List<Parameter> parameterList = mojoDescriptor.getParameters();
+        @SuppressWarnings( "unchecked" ) List<Parameter> parameterList = mojoDescriptor.getParameters();
```
rfscholte: Is the suppressWarning still required?

maven-script/maven-plugin-tools-model/src/main/java/org/apache/maven/tools/plugin/extractor/model/PluginMetadataParser.java
```diff
@@ -71,36 +67,25 @@
 
             if ( mojos != null && !mojos.isEmpty() )
```
rfscholte: `&& !mojos.isEmpty()` can be removed, let the for-loop handle this.

maven-plugin-plugin/pom.xml
```diff
@@ -390,6 +390,25 @@
         </plugins>
       </build>
     </profile>
+    <profile>
```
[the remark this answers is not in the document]
olamy: uhm not so sure if the argument is propagated.... see https://builds.apache.org/blue/organizations/jenkins/maven-box%2Fmaven-plugin-tools/detail/jdk11_asm7/7/pipeline/32

maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/PluginXdocGenerator.java
olamy: I guess not but I didn't upgrade to maven 3 :)

PR #17, opened 2019-08-31
Prefer try-with-resources for file error handling

gaul:
Closing an `InputStream` in a finally block can mask exceptions thrown
in the `try` or `catch` blocks.  Found via error-prone.  Reference:

https://errorprone.info/bugpattern/Finally

### DIFF
--- a/maven-plugin-tools-generators/src/main/resources/help-class-source.vm
+++ b/maven-plugin-tools-generators/src/main/resources/help-class-source.vm
@@ -114,10 +114,8 @@ public class HelpMojo
         throws MojoExecutionException
     {
         getLog().debug( "load plugin-help.xml: " + PLUGIN_HELP_PATH );
-        InputStream is = null;
-        try
+        try ( InputStream is = getClass().getResourceAsStream( PLUGIN_HELP_PATH ) )
         {
-            is = getClass().getResourceAsStream( PLUGIN_HELP_PATH );
             DocumentBuilderFactory dbFactory = DocumentBuilderFactory.newInstance();
             DocumentBuilder dBuilder = dbFactory.newDocumentBuilder();
             return dBuilder.parse( is );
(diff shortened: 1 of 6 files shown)


rfscholte:
Maybe this is a good moment to update the Java requirements and use try-with-resources

Tibor17:
@gaul Try to rewrite this title and try to find where we are missing Java 7 features. We ASF developers are not so fast to update all 100 Maven projects, so it means we need some help from you contributors.
Then pls run the build `mvn -P run-its verify` to make sure all changes are alright.

maven-plugin-tools-generators/src/main/resources/help-class-source.vm
```diff
@@ -144,7 +144,7 @@ public class HelpMojo
                 }
                 catch ( IOException e )
                 {
-                    throw new MojoExecutionException( e.getMessage(), e );
+                    getLog().warn( "Could not close " + PLUGIN_HELP_PATH + ": " + e.getMessage() );
```
eolivelli: We should use a  try-with-resources block, the exception will be eventually added as 'suppressed'

Tibor17: @eolivelli 
If @gaul would fix all Java 7 features then we should create a new Jira ticket as we did it in other projects.

gaul: I would like to let maven plugins use error-prone with the fewest suppressions.  Moving the entire project to Java 7 is one way to do this although I don't know if that aligns with the project's goals.

Tibor17: It perfectly fits to our goals. ;-)
We set the Javac version 1.7 and we were updating the code to J7.
Later we will do the same for J8 - inner classes/Lambda, I/O, loops with method chaining, use of the class `Optional`, etc.

gaul: Done.

gaul:
Superseded by #18.

PR #25, opened 2020-06-30
remove unneeded dependency

elharo:


michael-o:
Did you run mvn `dependency:analyze`?

elharo:
That's the command I was looking for. Thanks. I just happened to notice this one because it's an uncommon one. Running dependency:analyze doesn't show a problem here, though there are a lot of both unused and missing dependencies in other submodules.

michael-o:
After the run, are we safe to remove this dep?

elharo:
Yes, though there's a lot of cleanup to be done for other dependencies in other submodules. That might help with updating some of the tests for various plugins that depend on this project to Maven 3.1.

PR #29, opened 2020-08-10
(doc) added "help" goal; goal number corrected

Archangel1C:
There were 4 goals listed, but it said "has six goals". Also the "help" goal was not listed.
Added the "help" goal and corrected the number from six to five.


pzygielo:
> and corrected the number from six to five.

Somehow I can't find that correction...

Archangel1C:
@pzygielo Damn, you're right. I did it twice and the second time, I forgot to change the "six"...

Is there a way to amend this or should I just open a new pull request?

pzygielo:
> Is there a way to amend this or should I just open a new pull request?

No idea.
But should you decide for any way to do it - may I suggest tiny change:

> corrected the number from six to five.

IMO it would be future-proof to *change **six** to **several*** to have 

>    The Plugin Plugin has several goals:

[Dependency plugin](http://maven.apache.org/plugins/maven-dependency-plugin/) uses that.
No added/removed goal requires change of that sentence :grin:

PR #30, opened 2020-11-18
[MPLUGIN-367] Improve @Execute(lifecycle...) docs

mthmulders:


maven-plugin-annotations/src/main/java/org/apache/maven/plugins/annotations/Execute.java
```diff
@@ -51,7 +51,8 @@
     String goal() default "";
 
     /**
-     * lifecycle id to fork.
+     * lifecycle id of the phase to fork. Only valid in combination with {@link #phase()}. If not specified, Maven will
```
MartinKanters, line 54: Great improvement already, but I think "lifecycle id of the phase to fork. " is still a bit too vague.
Perhaps something like: "lifecycle id specifying which lifecycle the forked phase will run in."

And should we refer to lifecycles.xml? http://maven.apache.org/ref/3.2.5/maven-plugin-api/lifecycle-mappings.html

mthmulders: I chose slightly different words but added the suggested documentation reference.

MartinKanters: Looking good, but it seems I gave you the http link, sorry! Let's go for the https link instead (doesn't matter much, but looks better).
https://maven.apache.org/ref/3.6.3/maven-plugin-api/lifecycle-mappings.html

MartinKanters (approved):
LGTM! Very useful to have this doc cleared up 👍

rfscholte (approved):

PR #31, opened 2020-11-18
[MPLUGIN-368] Improve @Execute(goal...) docs

mthmulders:


maven-plugin-annotations/src/main/java/org/apache/maven/plugins/annotations/Execute.java
```diff
@@ -45,7 +45,8 @@
     LifecyclePhase phase() default LifecyclePhase.NONE;
 
     /**
-     * goal to fork. Note that specifying a phase overrides specifying a goal.
+     * goal to fork. Note that specifying a phase overrides specifying a goal. The goal must be a goal of the same
```
rfscholte, line 48: > ...the goal must be a goal
Looks a bit confusing. What I think you want to say is:
`The specified <code>goal</code> must be another existing goal of the same plugin`
(and yes, the code-tags are intended for javadoc)

mthmulders: Yes, that's absolutely an improvement. Thanks for the suggestion!

MartinKanters (approved):
LGTM!

michael-o (approved):

rfscholte (approved):

PR #32, opened 2020-11-22
[MPLUGIN-369] - Upgrade ASM Api version to ASM9 in visitors

vovkss:
This fixes the "Record requires ASM8" problem when Records are used in a Mojo.


slachiewicz (approved):

slachiewicz:
Thx merged to master

PR #33, opened 2021-04-28
[MPLUGIN-372] Descriptor mojo fails if super class is provided scope

cstamas:
https://issues.apache.org/jira/browse/MPLUGIN-372

Changes:
- Bug: adjust descriptor mojo resolution scope to include provided scope as well
- drop use of deprecated `project.getRuntimeDependencies()`
- Sanitized generator and report mojos (injects both project and project.artifacts, just one is completely enough)
- GeneratorUtils using Artifacts not Dependencies (as they were not needed)
- JavaAnnotationsMojoDescriptorExtractor error reporting: report failed class at least

### DIFF
--- a/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
+++ b/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
@@ -262,12 +262,7 @@ public abstract class AbstractGeneratorMojo
         {
             throw new MojoExecutionException( "Error writing plugin descriptor", e );
         }
-        catch ( InvalidPluginDescriptorException e )
-        {
-            throw new MojoExecutionException( "Error extracting plugin descriptor: \'" + e.getLocalizedMessage() + "\'",
-                                              e );
-        }
-        catch ( ExtractionException e )
+        catch ( InvalidPluginDescriptorException | ExtractionException e )
         {
             throw new MojoExecutionException( "Error extracting plugin descriptor: \'" + e.getLocalizedMessage() + "\'",
                                               e );
--- a/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
+++ b/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
@@ -127,14 +127,6 @@ public abstract class AbstractGeneratorMojo
     protected boolean skip;
 
     /**
-     * The set of dependencies for the current project
-     *
-     * @since 3.0
-     */
-    @Parameter( defaultValue = "${project.artifacts}", required = true, readonly = true )
-    protected Set<Artifact> dependencies;
-    
-    /**
      * Specify the dependencies as {@code groupId:artifactId} containing (abstract) Mojos, to filter
      * dependencies scanned at runtime and focus on dependencies that are really useful to Mojo analysis.
      * By default, the value is {@code null} and all dependencies are scanned (as before this parameter was added).
@@ -250,7 +242,7 @@ public abstract class AbstractGeneratorMojo
 
         try
         {
-            List<ComponentDependency> deps = GeneratorUtils.toComponentDependencies( project.getRuntimeDependencies() );
+            List<ComponentDependency> deps = GeneratorUtils.toComponentDependencies( project.getArtifacts() );
             pluginDescriptor.setDependencies( deps );
 
             PluginToolsRequest request = new DefaultPluginToolsRequest( project, pluginDescriptor );
@@ -313,7 +305,7 @@ public abstract class AbstractGeneratorMojo
         Set<Artifact> filteredDependencies;
         if ( mojoDependencies == null )
         {
-            filteredDependencies = dependencies;
+            filteredDependencies = new LinkedHashSet<>( project.getArtifacts() );
         }
         else if ( mojoDependencies.size() == 0 )
         {
@@ -325,7 +317,7 @@ public abstract class AbstractGeneratorMojo
             
             ArtifactFilter filter = new IncludesArtifactFilter( mojoDependencies );
 
-            for ( Artifact artifact : dependencies )
+            for ( Artifact artifact : project.getArtifacts() )
             {
                 if ( filter.include( artifact ) )
                 {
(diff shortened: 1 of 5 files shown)


cstamas:
This change should be addressed, as it is crucial to put plugins maven deps to provided.

rfscholte:
By just looking at the code I don't see weird things. This ticket confirms my feeling that there was something wrong, but I can't see the impact yet. I guess you can see only the difference if MNG-7097 is ready.

michael-o (commented):
I really don't have an opion on this. Robert is the expect here.

cstamas:
@rfscholte ping.

Some explanation: MPLUGIN-372 (this PR) and MNG-7097 are not about same thing:
- this PR fixes an issue about plugin tooling that currently prevent us using `provided` scope for maven artifacts. Is needed to implement MPLUGIN-370. So is for "future" (yet to be released) plugins.
- OTOH, MNG-7097 is for "everything else" (and old plugins, not having MPLUGIN-370 implemented), but MNG-7097 needs more changes to allow cases where "all is needed", see MNG-5783... am thinking to extend Mojo annotation and make Mojos like these "ask" for full (MNG-7097 disabled) resolution?

cstamas:
Or it could be other way around: add a Mojo annotation that plugin is "happy" with MNG-7097 enabled? This seems a bit counter-productive, as IMHO 99% of plugins are happy with it, only some exceptions like cobertura (or other crafting CP and spawning JVM may suffer from it).

rfscholte:
To me cobertura shouldn't block us for implementing this improvement.

maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/GeneratorUtils.java
```diff
@@ -120,14 +121,14 @@ public static void element( XMLWriter w, String name, String value, boolean asTe
     }
     
     /**
-     * @param dependencies not null list of <code>Dependency</code>
+     * @param dependencies not null collection of <code>Artifact</code>
      * @return list of component dependencies
      */
-    public static List<ComponentDependency> toComponentDependencies( List<Dependency> dependencies )
+    public static List<ComponentDependency> toComponentDependencies( Collection<Artifact> dependencies )
```
rfscholte, line 127: rename parameter to `artifacts`

maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/GeneratorUtils.java
```diff
@@ -120,14 +121,14 @@ public static void element( XMLWriter w, String name, String value, boolean asTe
...
     {
         List<ComponentDependency> componentDeps = new LinkedList<>();
 
-        for ( Dependency dependency : dependencies )
+        for ( Artifact dependency : dependencies )
```
rfscholte, line 131: rename variable to `artifact`

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/PluginReport.java
```diff
@@ -316,13 +307,13 @@ private PluginDescriptor extractPluginDescriptor()
...
         {
-            List<ComponentDependency> deps = GeneratorUtils.toComponentDependencies( project.getRuntimeDependencies() );
+            List<ComponentDependency> deps = GeneratorUtils.toComponentDependencies( project.getArtifacts() );
             pluginDescriptor.setDependencies( deps );
 
             PluginToolsRequest request = new DefaultPluginToolsRequest( project, pluginDescriptor );
             request.setEncoding( encoding );
             request.setSkipErrorNoDescriptorsFound( true );
-            request.setDependencies( dependencies );
+            request.setDependencies( new LinkedHashSet<>( project.getArtifacts() ) );
```
rfscholte, line 316: Are we adding artifacts or dependencies?

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
```diff
@@ -313,7 +305,7 @@ static String getDefaultGoalPrefix( MavenProject project )
         Set<Artifact> filteredDependencies;
         if ( mojoDependencies == null )
         {
-            filteredDependencies = dependencies;
+            filteredDependencies = new LinkedHashSet<>( project.getArtifacts() );
```
rfscholte, line 308: dependencies or artifacts?

rfscholte (changes requested):
In general we need to be clear if we are talking about dependencies or artifacts.
http://maven.apache.org/shared/maven-artifact-transfer/comparison.html tries to explain the difference.

maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/GeneratorUtils.java
cstamas: fixed

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/PluginReport.java
cstamas: fixed

maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/GeneratorUtils.java
cstamas: fixed

cstamas:
@rfscholte for the rest of your comments: well, as it can be seen, these are all uses of `org.apache.maven.project.MavenProject#getArtifacts` method, which in turn has JavaDoc saying "All dependencies that this project has...". Also, old (removed) code was calling it "dependencies", so I did not change that. The util class is aligned though.

cstamas:
ping

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
rfscholte: Renaming `filteredDependencies` to `filteredArtifacts` should be sufficient, right?

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/PluginReport.java
rfscholte: There's nothing we can do here anymore. Could be a different ticket to fix this inconsistency.

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
cstamas: done

rfscholte (approved):

cstamas:
merged as 8f354439765afdbf6ce166d299791117576e90f2

PR #34, opened 2021-04-30
Add GH actions

cstamas:
As they provide nice (and quick) feedback about issues
and problems with the build.

This action is simple action copied from resolver (that is based on maven action, but simplified).

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,49 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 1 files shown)


rfscholte:
I only care about https://builds.apache.org/job/maven/job/maven-box/job/maven-plugin-tools/ and to keep https://ci-builds.apache.org/job/Maven/job/dist-tool-plugin/job/master/site/dist-tool-master-jobs.html clean.

cstamas:
Of course, CI is the definitive, GH actions as you see does not install even, they are here just to run the "chores" like checkstyle and filter out obvious things.

MartinKanters:
@cstamas I agree, super useful for non-committer contributors as well.

PR #45, opened 2021-10-18
Weed out ITs

cstamas:
maven-plugin-api 2.0? Really?


michael-o:
> 
> 
> maven-plugin-api 2.0? Really?

Because we can!

PR #46, opened 2021-10-22
[MPLUGIN-378] Group extractors and improve their selection, deprecate javadoc extractor

cstamas:
Fix the execution order of extractors and extend their API to allow their deprecation
and based on deprecation we can log or fail the build even (later).

Extend extractors iface with following fields:
* getName() -- just to simplify their handling in code (no need to hold on map).
* isDeprecated() -- IMO, we should deprecate Javadoc mojo descriptor extraction, and
  if used, Maven should WARN to migrate to Annotations.
* getGroupKey() -- this defines "extractor" groups and ordering within it

Currently we have two extractors in same group "java": the annotation extractor
and javadoc extractor (BOTH use qdox as well). Their run order is _undefined currently_.

This proposal implements following changes:
* extractors are SORTED by groupKey before being used (so defined execution ordering)
* extractors of same group are ordered (annotations then javadoc)
* javadoc extractor is DEPRECATED (isDeprecated returns true)
* MojoScanner is modified, that IF a deprecated extractor is on turn, and it discovers 
  components, it will WARN. If Java extractors, it will warn to use Mojo Annotations.

---

https://issues.apache.org/jira/browse/MPLUGIN-378


maven-plugin-tools-api/src/main/java/org/apache/maven/tools/plugin/extractor/GroupKey.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private final String group;
+
+    private final int order;
+
+    public GroupKey( String group, int order )
+    {
+        if ( group == null )
+        {
+            throw new NullPointerException( "GroupKey.group null" );
```
michael-o, line 46: `Objects#requireNonNull`?

michael-o (commented):
Is there no way to detect `@Deprecated` instead of introducing our own `#isDeprecated()`?

cstamas:
> Is there no way to detect `@Deprecated` instead of introducing our own `#isDeprecated()`?

You mean to detect is extractor deprecated?

maven-plugin-tools-api/src/main/java/org/apache/maven/tools/plugin/extractor/GroupKey.java
cstamas: This is Java7 project, but yes, we may want to up it to Java8 for 3.7.0

gnodet:
@slachiewicz @slawekjaranowski @cstamas so what's the status of this PR ? I've merged master, but should that be integrated into the upcoming 3.7.0 release or delayed ?

slawekjaranowski:
For me can be merged for 3.7.0 especially with deprecating JavaDoc extractor.

jira-importer:
Resolve #691

PR #51, opened 2021-11-20
[MPLUGIN-380] - Always include threadSafety in report

pzygielo:
Follow-up to https://github.com/apache/maven-scm/pull/107#issuecomment-974704356

I do not speak languages, so commits for localization would be appreciated.

### DIFF
--- a/maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/PluginXdocGenerator.java
+++ b/maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/PluginXdocGenerator.java
@@ -314,13 +314,12 @@ public class PluginXdocGenerator
             }
         }
 
-        if ( mojoDescriptor.isThreadSafe() )
-        {
-            addedUl = addUl( w, addedUl );
-            w.startElement( "li" );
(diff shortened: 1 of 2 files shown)


maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/PluginXdocGenerator.java
```diff
@@ -312,17 +312,15 @@ private void writeGoalAttributes( MojoDescriptor mojoDescriptor, XMLWriter w )
...
-                w.startElement( "li" );
-                w.writeMarkup( getString( "pluginxdoc.mojodescriptor.threadSafe" ) );
-                w.endElement(); //li
-            }
-
         }
 
+        addedUl = addUl( w, addedUl );
+        w.startElement( "li" );
+        w.writeMarkup( getString( mojoDescriptor.isThreadSafe()
```
michael-o, line 319: Why did you replace extended with non-extended?

pzygielo: - Please check #50 first.

jira-importer:
Resolve #676

PR #52, opened 2021-11-28
[MPLUGIN-382] exclude dependency in provided scope from plugin descriptor

slawekjaranowski:
`plugin.xml` descriptor - should only contain dependencies in compile scope

### DIFF
--- a/maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/GeneratorUtils.java
+++ b/maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/GeneratorUtils.java
@@ -19,7 +19,10 @@ package org.apache.maven.tools.plugin.generator;
  * under the License.
  */
 
-import static java.nio.charset.StandardCharsets.UTF_8;
+import javax.swing.text.MutableAttributeSet;
+import javax.swing.text.html.HTML;
+import javax.swing.text.html.HTMLEditorKit;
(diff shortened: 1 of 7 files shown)


michael-o:
@cstamas Can you have a look?

maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/GeneratorUtils.java
```diff
@@ -122,14 +122,19 @@ public static void element( XMLWriter w, String name, String value, boolean asTe
...
-     * @return list of component dependencies
+     * @return list of component dependencies, only in compile scope
      */
     public static List<ComponentDependency> toComponentDependencies( Collection<Artifact> artifacts )
     {
         List<ComponentDependency> componentDeps = new LinkedList<>();
 
         for ( Artifact artifact : artifacts )
         {
+            if ( !Artifact.SCOPE_COMPILE.equals( artifact.getScope() ) )
```
cstamas, line 133: I would rather make `SCOPE = PROVIDED` (then continue). What about runtime scope? Just to be on safe side.

slawekjaranowski: fixed

cstamas (approved):

jira-importer:
Resolve #692

PR #56, opened 2021-12-17
[MPLUGIN-386] Lower the tone and limit the scope check to maven-plugin-api only

cstamas:
The warning was too chatty, a bit pushy, and also it
triggered on too broad condition (all of org.apache.maven groupIds).

This change makes following modifications:
* only maven-plugin-api is checked for (so JXR, Archiver will not trigger it)
* message is really one liner about best practice (if anyone curious, go after it anyway)

---

https://issues.apache.org/jira/browse/MPLUGIN-386

### DIFF
--- a/maven-plugin-plugin/src/it/mplugin-370-maven-deps-scope-bad/verify.groovy
+++ b/maven-plugin-plugin/src/it/mplugin-370-maven-deps-scope-bad/verify.groovy
@@ -19,4 +19,4 @@
 
 File buildLog = new File( basedir, "build.log" );
 assert buildLog.isFile()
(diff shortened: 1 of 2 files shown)


slachiewicz (approved):
Yes yes, also saw that earlier

michael-o (commented):
I don't understand this change. Why only the plugin API? Ignoring Archiver and JXR for now.

rmannibucau (approved):
approving since we can't do really better for now but will need to be revisited anyway since both (only this or all maven*) strategies are wrong and we need to cover the actual stable maven API

cstamas:
Superseded by https://github.com/apache/maven-plugin-tools/pull/58

jira-importer:
Resolve #686

PR #58, opened 2021-12-17
[MPLUGIN-386] Exclude maven-archiver and maven-jxr from warning

cstamas:
Those are not Maven Core components, they are merely in
wrong groupId. Extended ITs to use maven-jxr as well.

### DIFF
--- a/maven-plugin-plugin/src/it/mplugin-370-maven-deps-scope-bad/pom.xml
+++ b/maven-plugin-plugin/src/it/mplugin-370-maven-deps-scope-bad/pom.xml
@@ -40,12 +40,6 @@ under the License.
   <dependencies>
     <dependency>
       <groupId>org.apache.maven</groupId>
-      <artifactId>maven-jxr</artifactId>
-      <version>3.1.1</version>
-      <scope>compile</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.maven</groupId>
       <artifactId>maven-plugin-api</artifactId>
       <version>@mavenVersion@</version>
       <scope>compile</scope>
--- a/maven-plugin-plugin/src/it/mplugin-370-maven-deps-scope-bad/pom.xml
+++ b/maven-plugin-plugin/src/it/mplugin-370-maven-deps-scope-bad/pom.xml
@@ -40,6 +40,12 @@ under the License.
   <dependencies>
     <dependency>
       <groupId>org.apache.maven</groupId>
+      <artifactId>maven-jxr</artifactId>
+      <version>3.1.1</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
       <artifactId>maven-plugin-api</artifactId>
       <version>@mavenVersion@</version>
       <scope>compile</scope>
--- a/maven-plugin-plugin/src/it/mplugin-370-maven-deps-scope-good/pom.xml
+++ b/maven-plugin-plugin/src/it/mplugin-370-maven-deps-scope-good/pom.xml
@@ -40,6 +40,12 @@ under the License.
   <dependencies>
     <dependency>
       <groupId>org.apache.maven</groupId>
+      <artifactId>maven-jxr</artifactId>
+      <version>3.1.1</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
       <artifactId>maven-plugin-api</artifactId>
       <version>@mavenVersion@</version>
       <scope>provided</scope>
--- a/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
+++ b/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
@@ -327,8 +327,11 @@ public abstract class AbstractGeneratorMojo
 
         for ( Artifact dependency : project.getArtifacts() )
         {
+            // maven-archiver and maven-jxr are really in wrong package
             if ( "org.apache.maven".equals( dependency.getGroupId() )
                 && dependency.getArtifactId().startsWith( "maven-" )
+                && !dependency.getArtifactId().equals( "maven-archiver" )
+                && !dependency.getArtifactId().equals( "maven-jxr" )
                 && !Artifact.SCOPE_PROVIDED.equals( dependency.getScope() ) )
             {
                 wrongScopedDependencies.add( dependency );
--- a/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
+++ b/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
@@ -327,7 +327,7 @@ public abstract class AbstractGeneratorMojo
 
         for ( Artifact dependency : project.getArtifacts() )
         {
-            // maven-archiver and maven-jxr are really in wrong package
+            // maven-archiver and maven-jxr are really in wrong groupId
             if ( "org.apache.maven".equals( dependency.getGroupId() )
                 && dependency.getArtifactId().startsWith( "maven-" )
                 && !dependency.getArtifactId().equals( "maven-archiver" )
--- a/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
+++ b/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
@@ -185,8 +185,8 @@ public abstract class AbstractGeneratorMojo
     private List<String> expectedProvidedScopeGroupIds = Collections.singletonList( "org.apache.maven" );
 
     /**
-     * List of {@code G:A} strings of artifact coordinates that are to be excluded from "expected provided scope" check.
-     * Default value: {@code ["org.apache.maven:maven-archiver", "org.apache.maven:maven-jxr"]}.
+     * List of {@code groupID:artifactID} strings of artifact coordinates that are to be excluded from "expected
+     * provided scope" check. Default value: {@code ["org.apache.maven:maven-archiver", "org.apache.maven:maven-jxr"]}.
      *
      * @since 3.6.3
      */
--- a/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
+++ b/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
@@ -185,7 +185,7 @@ public abstract class AbstractGeneratorMojo
     private List<String> expectedProvidedScopeGroupIds = Collections.singletonList( "org.apache.maven" );
 
     /**
-     * List of {@code groupID:artifactID} strings of artifact coordinates that are to be excluded from "expected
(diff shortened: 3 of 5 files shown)


michael-o (approved):

rmannibucau:
Can we really do it outside maven plugin api subgraph? Dont think so today :(

michael-o:
> Can we really do it outside maven plugin api subgraph? Dont think so today :(

WDYM?

rmannibucau:
@michael-o except maven-plugin-api which is intentionally exposed rest of maven sub-artifacts are random and either internals or public API so none can be assumed being provided so this sounds like a lot of false positive to me. What about not warning for all but the plugin api and maybe maven-core?

kwin:
@rmannibucau Which other false positives do you foresee? I am actually thankful for the warnings for "maven-artifact" and "maven-model" (those are API IMHO since forever)

rmannibucau:
Maven comes OOTB with:

````
apache-maven-3.8.4\lib\maven-artifact-3.8.4.jar
apache-maven-3.8.4\lib\maven-builder-support-3.8.4.jar
apache-maven-3.8.4\lib\maven-compat-3.8.4.jar
apache-maven-3.8.4\lib\maven-core-3.8.4.jar
apache-maven-3.8.4\lib\maven-embedder-3.8.4.jar
apache-maven-3.8.4\lib\maven-model-3.8.4.jar
apache-maven-3.8.4\lib\maven-model-builder-3.8.4.jar
apache-maven-3.8.4\lib\maven-plugin-api-3.8.4.jar
apache-maven-3.8.4\lib\maven-repository-metadata-3.8.4.jar
apache-maven-3.8.4\lib\maven-resolver-api-1.6.3.jar
apache-maven-3.8.4\lib\maven-resolver-connector-basic-1.6.3.jar
apache-maven-3.8.4\lib\maven-resolver-impl-1.6.3.jar
apache-maven-3.8.4\lib\maven-resolver-provider-3.8.4.jar
apache-maven-3.8.4\lib\maven-resolver-spi-1.6.3.jar
apache-maven-3.8.4\lib\maven-resolver-transport-wagon-1.6.3.jar
apache-maven-3.8.4\lib\maven-resolver-util-1.6.3.jar
apache-maven-3.8.4\lib\maven-settings-3.8.4.jar
apache-maven-3.8.4\lib\maven-settings-builder-3.8.4.jar
apache-maven-3.8.4\lib\maven-shared-utils-3.3.4.jar
apache-maven-3.8.4\lib\maven-slf4j-provider-3.8.4.jar
````

From there, maven-plugin-api and maven-model are public, part (part is important there) of maven-core is too and rest is fully undefined or private as of today - not 100% sure for artifact since it changed a bit but agree it used to leak so I assume it is okish to include it. This is only for built-in maven artifacts and not maven ecosystem so guess we can only whitelist the 2-3 dependencies and not include all and exclude a few in terms of impl.

File: maven-plugin-plugin/src/it/mplugin-370-maven-deps-scope-bad/pom.xml [hunk 1]
```diff
@@ -38,6 +38,12 @@ under the License.
   </properties>
 
   <dependencies>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-jxr</artifactId>
+      <version>3.1.1</version>
+      <scope>compile</scope>
+    </dependency>
```
slawekjaranowski, line 46: There are no assertions for it ... so here is no need

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
```diff
@@ -327,8 +327,11 @@ static String getDefaultGoalPrefix( MavenProject project )
 
         for ( Artifact dependency : project.getArtifacts() )
         {
+            // maven-archiver and maven-jxr are really in wrong groupId
             if ( "org.apache.maven".equals( dependency.getGroupId() )
                 && dependency.getArtifactId().startsWith( "maven-" )
+                && !dependency.getArtifactId().equals( "maven-archiver" )
+                && !dependency.getArtifactId().equals( "maven-jxr" )
```
slawekjaranowski, line 334: Maybe put excludes names in some list to simplify expression ... 
will be easier add next exclude

maven-plugin-plugin/src/it/mplugin-370-maven-deps-scope-bad/pom.xml
cstamas: There are assertions to NOT have warnings, so to say, that these "excludes" are obeyed (maven-archiver and maven-jxr).

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
cstamas: fixed

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
```diff
@@ -163,7 +164,36 @@
...
+    private List<String> expectedProvidedScopeGroupIds = Collections.singletonList( "org.apache.maven" );
+
+    /**
+     * List of {@code G:A} strings of artifact coordinates that are to be excluded from "expected provided scope" check.
+     * Default value: {@code ["org.apache.maven:maven-archiver", "org.apache.maven:maven-jxr"]}.
+     *
+     * @since 3.6.3
+     */
+    @Parameter
+    private List<String> expectedProvidedScopeExclusions = Arrays.asList(
```
michael-o, line 194: But those aren't expected to be in provided scope, are they? They are just int he wrong group id.

cstamas: Correct, hence they are excluded

michael-o: OK, thanks.

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
```diff
@@ -163,7 +164,36 @@
      * @since 3.3
      */
     @Parameter
-    protected List<String> packagingTypes = Arrays.asList( "maven-plugin" );
+    protected List<String> packagingTypes = Collections.singletonList( "maven-plugin" );
```
michael-o, line 167: Can this ever change?

cstamas: unsure, I just got rid of warning (one param to Arrays.asList).

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
```diff
@@ -163,7 +164,36 @@
...
+     * List of {@code groupId} strings of artifact coordinates that are expected to be in "provided" scope. Default
+     * value: {@code ["org.apache.maven"]}.
+     *
+     * @since 3.6.3
+     */
+    @Parameter
+    private List<String> expectedProvidedScopeGroupIds = Collections.singletonList( "org.apache.maven" );
+
+    /**
+     * List of {@code G:A} strings of artifact coordinates that are to be excluded from "expected provided scope" check.
```
michael-o, line 188: You should write out `G:A`

michael-o (changes requested):

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
michael-o: IF this will never change then make is static final string?

maven-plugin-plugin/src/it/mplugin-370-maven-deps-scope-bad/pom.xml
cstamas: Wrong: removed

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
cstamas: fixed

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
cstamas: in any case, that change is unrelated to this PR... Again, I just fixed a warning, but I can undo this change

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
```diff
@@ -185,8 +185,8 @@
     private List<String> expectedProvidedScopeGroupIds = Collections.singletonList( "org.apache.maven" );
 
     /**
-     * List of {@code G:A} strings of artifact coordinates that are to be excluded from "expected provided scope" check.
-     * Default value: {@code ["org.apache.maven:maven-archiver", "org.apache.maven:maven-jxr"]}.
+     * List of {@code groupID:artifactID} strings of artifact coordinates that are to be excluded from "expected
```
michael-o, line 188: Well, the field above days `groupId`, not `groupID` and the former is the term we use everywhere.

michael-o (changes requested):

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
michael-o: Yes, you are right. Let's keep your change. I have just noticed this and it felt awkward.

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
cstamas: Fixed

michael-o (approved):

maven-plugin-plugin/src/it/mplugin-370-maven-deps-scope-good/pom.xml
```diff
@@ -38,6 +38,12 @@ under the License.
   </properties>
 
   <dependencies>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-jxr</artifactId>
+      <version>3.1.1</version>
+      <scope>compile</scope>
+    </dependency>
```
slawekjaranowski, line 46: Should be also added `org.apache.maven:maven-archiver`  unless is present on dependency tree

cstamas: We have identified these two "cuckoo eggs" but am using only jxr in IT as maven-archiver pulls in a bunch of other deps, making IT run eons. I think for IT using jxr only is okay .

slawekjaranowski (approved):

jira-importer:
Resolve #690

PR #59, opened 2022-01-09
[MPLUGIN-384] restore compatibility with Maven 3 ecosystem

hboutemy:
https://issues.apache.org/jira/browse/MPLUGIN-384

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ out/
 .DS_Store
 
 .java-version
+.factorypath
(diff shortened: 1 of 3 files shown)


kwin (commented):
IIUC then two dependencies would expose the same fqcn (compat and m-p-p). Wouldn't that potentially cause trouble (ClassCastException) I think at least a hint in the javadoc is in order.

hboutemy:
I think that Maven class loading won't load the class from the plugin: it is taken from Maven core classloader (that contains compat), see https://maven.apache.org/guides/mini/guide-maven-classloading.html
but you're right that it deserves testing in real life to check that I'm not wrong (IIRC, I already saw such a class copy in another case, but I don't remember where)...

slawekjaranowski:
I checked with real environment, metadata to oss was deployed correctly.

slawekjaranowski (approved):

hboutemy:
thank you Slawomir

jira-importer:
Resolve #688

PR #66, opened 2022-01-26
Reshuffle plugin dependencies

cstamas:
Changes:
* use m-p-p 3.6.4 finally, so maven bits can be provided scope
* maven-reporting-impl is optional (so in case standalone mojo is used, it needs to be added to plugin dep)
* align use of reporting and doxia across modules (there were several versions used)

### DIFF
--- a/maven-plugin-plugin/pom.xml
+++ b/maven-plugin-plugin/pom.xml
@@ -42,17 +42,10 @@
   </prerequisites>
 
   <properties>
-    <doxiaVersion>1.11.1</doxiaVersion>
-    <doxia-sitetoolsVersion>1.11.1</doxia-sitetoolsVersion>
     <it.debug>true</it.debug>
   </properties>
 
   <dependencies>
-    <dependency>
-      <groupId>org.apache.maven</groupId>
-      <artifactId>maven-core</artifactId>
-      <scope>provided</scope>
-    </dependency>
     <dependency>
       <groupId>org.apache.maven.plugin-tools</groupId>
       <artifactId>maven-plugin-tools-api</artifactId>
@@ -102,36 +95,13 @@
     <dependency>
       <groupId>org.apache.maven.doxia</groupId>
       <artifactId>doxia-sink-api</artifactId>
-      <version>${doxiaVersion}</version>
-      <exclusions>
-        <exclusion>
-          <groupId>org.codehaus.plexus</groupId>
-          <artifactId>plexus-container-default</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.maven.doxia</groupId>
-      <artifactId>doxia-site-renderer</artifactId>
-      <version>${doxia-sitetoolsVersion}</version>
-      <exclusions>
-        <exclusion>
-          <groupId>org.codehaus.plexus</groupId>
-          <artifactId>plexus-container-default</artifactId>
-        </exclusion>
-        <exclusion>
-          <groupId>org.codehaus.plexus</groupId>
-          <artifactId>plexus-component-api</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
 
     <!-- Maven -->
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-plugin-api</artifactId>
-      <!-- TODO: make this provided once 3.6.2+ maven-plugin-plugin is in use -->
-      <scope>compile</scope>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>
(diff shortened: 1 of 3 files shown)


michael-o:
ATTENTION: Reporting plugin cannot have dependencies in the POM. I tripped over it recently.

cstamas:
> ATTENTION: Reporting plugin cannot have dependencies in the POM. I tripped over it recently.

? No, this is build/plugins/m-p-p plugin is that needs dep added IF you plan to use PluginReport mojo "standalone"

maven-plugin-plugin/pom.xml
```diff
@@ -270,8 +249,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-plugin-plugin</artifactId>
-        <version>3.6.0</version>
-        <!-- will use previous maven-plugin-plugin release to build current maven-plugin-plugin as configured in parent -->
+        <version>3.6.4</version>
```
slawekjaranowski, line 252: please also update version for reporting -> m-p-p

cstamas: and where do I do that?

slawekjaranowski: line 277

cstamas: tx, fixed

slawekjaranowski:
I see some inconsistency - cosmetics

- `mavenInvokerPluginVersion` - defined in root pom used in `maven-plugin-plugin/pom.xml`
- `mavenPluginPlugin.version` - defined and used `plugin-plugin/pom.xml`
- different format of properties for versions once with dot once without dot

maybe `pluginMangment` in project root pom?

cstamas:
@slawekjaranowski fixed then (moved property where used, renamed them to be camelCase).

maven-plugin-plugin/pom.xml
```diff
@@ -296,7 +276,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-plugin-plugin</artifactId>
-        <version>3.6.0</version><!-- cannot use ${mavenPluginToolsVersion} property because release plugin would try to update -->
```
slawekjaranowski, line 299: is it not valid?
will work during release?

cstamas: Um, no idea, I tend to avoid m-release-p, never used it (aside of Maven projects). But, if this (otherwise perfectly valid) construct "does not work" with m-release-p, isn't the release plugin that needs to be fixed? Or we want to sprinkle "workarounds" all around as "release plugin would try to update"?

cstamas: This is the origin of the comment: https://github.com/apache/maven-plugin-tools/commit/3b54fdbc289fbc33db2366c37579e13a0e087a08  the commit message to me looks like an issue description :smile: (and a very old one)

slawekjaranowski: https://issues.apache.org/jira/browse/MRELEASE-799
maybe it is a bug only for artifact dependency not for plugins versions ...

according to issue should be reproducible with
```
mvn release:prepare -DdryRun=true ...
```

cstamas: My take on this https://github.com/apache/maven-release/pull/100

slawekjaranowski (approved):

michael-o:
I would consider to move this part to a `maven-plugin-reporting-plugin`.

slawekjaranowski:
> I would consider to move this part to a `maven-plugin-reporting-plugin`. Similar to Surefire.

Idea is good, but it will be breaking change, we will have new artifact for reporting section.

michael-o:
> > I would consider to move this part to a `maven-plugin-reporting-plugin`. Similar to Surefire.
> 
> Idea is good, but it will be breaking change, we will have new artifact for reporting section.

Correct, that will be in 4.0.0 and that is fine for us. We can now already create it, mark as deprecated and add a relocation information. Remove in 4.0.0.

gnodet:
@slachiewicz @slawekjaranowski @cstamas so what's the status of this PR ? I've merged master, but should that be integrated into the upcoming 3.7.0 release or delayed ?

gnodet:
> @slachiewicz @slawekjaranowski @cstamas so what's the status of this PR ? I've merged master, but should that be integrated into the upcoming 3.7.0 release or delayed ?

I guess the discussion is leading towards splitting the plugins in 4.0 instead of merging this PR, right ?

michael-o:
> > @slachiewicz @slawekjaranowski @cstamas so what's the status of this PR ? I've merged master, but should that be integrated into the upcoming 3.7.0 release or delayed ?
> 
> @michael-o It seems the discussion is leading towards splitting the plugins in 4.0 instead of merging this PR, right ?

Correct, just like with Surefire Plugins.

slawekjaranowski (commented):
I agree, the better way will be move reports to dedicated plugin.
Even more  we have already have multimodule project.

gnodet:
Superseded by #147

jira-importer:
Resolve #697

PR #67, opened 2022-02-07
[MPLUGIN-390] Do not overwrite generate files with no content change

gnodet:


slawekjaranowski:
Some test would be usable for regression confirm.

maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/CachingOutputStream.java
```diff
@@ -0,0 +1,63 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Objects;
+
+/**
+ * Caching OutputStream to avoid overwriting a file with
+ * the same content.
+ */
+public class CachingOutputStream extends ByteArrayOutputStream
```
slawekjaranowski, line 34: I guess I've seen similar class in many place ... maybe can be put in some of share / common

slawekjaranowski (approved):

maven-plugin-plugin/src/it/mplugin-390/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+          <execution>
+            <id>touch</id>
+            <phase>generate-resources</phase>
+            <goals>
+              <goal>run</goal>
+            </goals>
+            <configuration>
+              <target>
+                <touch file="${project.build.directory}/touch.txt"/>
+                <sleep milliseconds="100"/>
```
slawekjaranowski, line 100: https://bugs.openjdk.java.net/browse/JDK-8177809

I don't remember where, but some time ago I have problem with lastModified .. so maybe a little longer

maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/CachingOutputStream.java
gnodet: I've introduced a similar one, but in modello (https://github.com/codehaus-plexus/modello/pull/116) and in plexus-containers (https://github.com/codehaus-plexus/plexus-containers/pull/46).  So only location would be in `plexus-utils` I suppose.  But the two classes in modello and plexus-containers are `Writer`s and this is the only `OutputStream`.

maven-plugin-plugin/src/it/mplugin-390/pom.xml
gnodet: Raised the sleep time to 1s

maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/CachingOutputStream.java
gnodet: Updated with the `CachingOutputStream` from plexus-utils.

michael-o (approved):
LGTM.

olamy (approved):

PR #72, opened 2022-02-24
[MPLUGIN-394] use XmlStreamReader to detect encoding instead of platform

hboutemy:


slawekjaranowski:
Is it possible to prepare a test which can confirm that works as expected?

slawekjaranowski:
please also rebase with current master base code.

slawekjaranowski:
#73 - next with tests 😄

hboutemy:
interesting bits with XmlStreamReader to be merged to #73

PR #74, opened 2022-02-25
[MPLUGIN-395] Upgrade Maven Reporting API/Impl to 3.1.0

michael-o:
Implicitly update MSITE to 3.11.0 and MPIR 3.2.2 for consistency with reporting.

This closes #74

### DIFF
--- a/maven-plugin-plugin/pom.xml
+++ b/maven-plugin-plugin/pom.xml
@@ -154,7 +154,7 @@
     <dependency>
       <groupId>org.apache.maven.reporting</groupId>
       <artifactId>maven-reporting-impl</artifactId>
-      <version>3.0.0</version>
+      <version>3.1.0</version>
     </dependency>
 
     <!-- plexus -->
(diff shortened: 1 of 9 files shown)


maven-plugin-plugin/pom.xml
```diff
@@ -350,7 +350,8 @@
               <settingsFile>src/it/settings.xml</settingsFile>
               <streamLogsOnFailures>true</streamLogsOnFailures>
               <filterProperties>
-                <sitePluginVersion>3.3</sitePluginVersion>
+                <sitePluginVersion>3.11.0</sitePluginVersion>
+                <projectInfoReportsPlugin>3.2.2</projectInfoReportsPlugin>
```
slawekjaranowski, line 354: Why not use the same version like in project?
Do we want later use another version in tests then in project?
I afraid that next we update version in project but not for tests.
And how to managed version if we have plugin version from parent?

michael-o: In most cases reactor plugins and IT plugins are decoupled. Look at other projects. Moreover, I did not want to rewrite the entire thing. You have valid concerns, but they need to be addressed on a global level for all components.

slawekjaranowski: ok, agreed

slachiewicz:
Maybe we should have separate upgrades of  one plugin and independent upgrade of it tests? To easier detect if plugin updates not affects old it tests?

michael-o:
> Maybe we should have separate upgrades of one plugin and independent upgrade of it tests? To easier detect if plugin updates not affects old it tests?

I totally agree with you and will explain why I did this in one go:
* Both API/Impl go hand on hand for Doxia 2 preparation
* When you update both reporting components ITs start to fail

Basically spilitting up is posslbe, but will leave commits in broken state that is why I have them together. In other projects I have upgraded ITs completely separately.

slachiewicz (approved):

PR #85, opened 2022-05-22
[MPLUGIN-389] Update level to Java 8

slawekjaranowski:
step by step 😄

### DIFF
--- a/maven-plugin-plugin/src/it/ant-reference-parameter/src/main/java/org/apache/maven/plugins/antWithRefs/ArtifactsTask.java
+++ b/maven-plugin-plugin/src/it/ant-reference-parameter/src/main/java/org/apache/maven/plugins/antWithRefs/ArtifactsTask.java
@@ -29,7 +29,7 @@ public class ArtifactsTask
     public void execute()
         throws BuildException
     {
-        log( String.valueOf( getProject().getReference( "artifacts" ) ) );
+        log( String.valueOf( ( Object ) getProject().getReference( "artifacts" ) ) );
     }
 
 }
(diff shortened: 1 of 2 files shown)


maven-plugin-plugin/src/it/ant-reference-parameter/src/main/java/org/apache/maven/plugins/antWithRefs/ArtifactsTask.java
```diff
@@ -29,7 +29,7 @@
     public void execute()
         throws BuildException
     {
-        log( String.valueOf( getProject().getReference( "artifacts" ) ) );
+        log( String.valueOf( ( Object ) getProject().getReference( "artifacts" ) ) );
```
michael-o, line 32: The reference is never null?

slawekjaranowski: Without such cast we will have exception:

```
Caused by: java.lang.ClassCastException: java.util.LinkedHashSet cannot be cast to [C
    at org.apache.maven.plugins.antWithRefs.ArtifactsTask.execute (ArtifactsTask.java:32)
    at org.apache.tools.ant.UnknownElement.execute (UnknownElement.java:299)
    at sun.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
```

As scripting Mojo are deprecated I didn't investigate it, simply did in such way to make test pass.

michael-o: Yes, true. That is fully acceptable!

michael-o (approved):

jira-importer:
Resolve #698

PR #114, opened 2022-07-04
[MPLUGIN-404] Set JDK prerequisite preferably from release parameter or

kwin:
maven.compiler.release property


jira-importer:
Resolve #712

PR #117, opened 2022-07-08
[MPLUGIN-441] Support for the new maven4 api

gnodet:
This PR includes the following:
 * raise the minimum JDK to 1.8
 * provide _additional_ support for the new maven4 api using maven 4.0.0-alpha-2
 
 The plugin still supports maven 3 plugins so is compatible with all the existing plugins.
The new support is provided by the annotation scanner which now supports the new package `org.apache.maven.api.plugin` in addition to the old one.  The main difference is that if the new package is detected, a flag will be set on the mojo descriptor so that maven can know that the mojo is a v4 mojo and behave accordingly.


maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/scanner/MojoAnnotationsScanner.java
```diff
@@ -37,13 +37,21 @@
 {
     String ROLE = MojoAnnotationsScanner.class.getName();
 
+    String V4_API_PLUGIN_PACKAGE = "org.apache.maven.api.plugin";
```
kwin, line 40: Maybe a bit late but why are the annotations part of maven-api-core? IMHO those are used exclusively at build time (not at run time) by m-plugin-tools only to build the descriptor, so why defining them globally?
Also why can't the old annotations be extended with just that a boolean flag for v4Api?

gnodet: > Maybe a bit late but why are the annotations part of maven-api-core? IMHO those are used exclusively at build time (not at run time) by m-plugin-tools only to build the descriptor, so why defining them globally?

The plugin annotations are the first classes needed for people that consume the maven api.  It's part of the API and I would not consider the fact that they are only used at build time (that's an implementation detail of how plugin works).  That is also related to the fact that the `MojoDescriptor` is currently not exposed in the api.  So there's a small change in semantics, as I consider the annotations as _the api_ for plugin authors, while the descriptor is kept an implementation detail (which is good imho as it inherit plexus classes which should be hidden).
Some attempts have been made over the years to define plugins in other languages such as the scripting stuff, but it's being deprecated and will be removed soon I think.  I think that's an indication that (nearly) all plugins are simply written using the provided annotations, so they _de facto_ constitute the mojo API.

> Also why can't the old annotations be extended with just that a boolean flag for v4Api?

They could.  Bur for the reason above, I think they should belong to the "_maven api_" which I think is best kept in a single location rather than splitted across various modules and projects.  Now, people can simply depend on the `maven-api-core` jar and have a clear view of all the maven api they can use.

kwin: > the annotations as the api for plugin authors

Agree, when they use the m-plugin-plugin for building their plugin (which is likely but not a must).
Still mixing runtime API and build time API in the same artifact is confusing to me.

> while the descriptor is kept an implementation detail (which is good imho as it inherit plexus classes which should be hidden).

Hmm, let us call it SPI at least as it defines the contract between Maven and its Plugins.

While I agree with most of your statements I still don't understand why those annotations need to be part of Maven rather than part of the (anyhow used) m-plugin tools. They are exclusively used there!
Also they inflate/pollute the Maven classpath for no reason as they are `RetentionPolicy.CLASS`. Mixing those different levels of API in a single artifact is not a very good design IMHO.

gnodet: I disagree that the fact they are not retained at runtime implies they should be in a different jar.  Should we change the retention in the new api ? ;-) 

More seriously, I agree they're not used at runtime, but I think from a user facing pov, they are the API for mojos, at the same level than the `Mojo` interface. The fact they are interpreted at compile time and translated into an xml is mainly an implementation detail for the average user. 
I think having a monolithic jar for the API  makes it way easier to just depend on the `maven-core-api` without having to look for additional jars, depending if you want to do simple mojos, some access to the local repository, some custom dependency collection or whatever part which has been moved to external components.  We don't even know which ones are deprecated and which ones should be used (for dependency collection / resolution).  With a single jar, everything coming from this jar (and it's dependencies) is part of the new API, everything else isn't.

maven-plugin-tools-generators/src/main/resources/help-class-source.vm
```diff
@@ -1,13 +1,20 @@
+#if ( $useMaven4Api )
+#set ( $mojoException = "MojoException" )
+#set ( $logger = "logger" )
+#else
+#set ( $mojoException = "MojoExecutionException" )
+#set ( $logger = "getLog()" )
+#end
```
slawekjaranowski, line 7: Maybe separate file for v3 and v4.
in V4 we can drop javadoc tags

gnodet: Done, and rebased / squashed.

maven-plugin-tools-api/src/main/java/org/apache/maven/tools/plugin/util/PluginUtils.java
```diff
@@ -63,7 +63,14 @@ private PluginUtils()
...
         mavenComponents.put( "org.apache.maven.plugin.descriptor.PluginDescriptor", "${plugin}" );
         mavenComponents.put( "org.apache.maven.settings.Settings", "${settings}" );
-        
+
+        mavenComponents.put( "org.apache.maven.api.Session", "${session}" );
+        mavenComponents.put( "org.apache.maven.api.Project", "${project}" );
+        mavenComponents.put( "org.apache.maven.api.MojoExecution", "${mojoExecution}" );
+        // TODO: apiv4: add PluginDescriptor to the api ?
+        //mavenComponents.put( "org.apache.maven.api.descriptor.PluginDescriptor", "${plugin}" );
+        mavenComponents.put( "org.apache.maven.api.settings.Settings", "${settings}" );
```
slawekjaranowski, line 72: I'm not sure about it ...

maven-plugin-plugin/pom.xml
```diff
@@ -38,11 +38,12 @@
   </description>
 
   <prerequisites>
-    <maven>${mavenVersion}</maven>
+    <maven>3.2.5</maven>
```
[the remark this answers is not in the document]
gnodet: Yes, that comes from when the PR had switche to mvn 4.

maven-plugin-tools-api/src/main/java/org/apache/maven/tools/plugin/util/PluginUtils.java
slawekjaranowski: As I remember `@Component Session session` is translated to  `@Parameter( property = "${session}" )` - to check

gnodet: Yes, and the map is here to do that.  So if I remove the population of the `mavenComponents` with objects from the v4 api, this won't work anymore.  The IT actually tests that.

gnodet: Also, I don't think the use of `@Parameter` or `@Component` should be dictated by the implementation.  For example, the maven session is available as both a parameter injected by the plugin manager and a sisu bean inside the `@SessionScope`.  Also `@Component` relates to the sisu components, but this does not have to be made explicit.
I would personally favor the `@Component` approach for everything that does not come from the configuration.

pom.xml
```diff
@@ -29,7 +29,7 @@
 
   <groupId>org.apache.maven.plugin-tools</groupId>
   <artifactId>maven-plugin-tools</artifactId>
-  <version>3.7.1-SNAPSHOT</version>
+  <version>4.0.0-SNAPSHOT</version>
```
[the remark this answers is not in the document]
gnodet: I've reverted the change to 4.0.0-SNAPSHOT

maven-plugin-tools-api/src/main/java/org/apache/maven/tools/plugin/util/PluginUtils.java
cstamas: Me too.

maven-plugin-plugin/src/it/settings.xml
```diff
@@ -39,9 +39,11 @@ under the License.
           <url>@localRepositoryUrl@</url>
           <releases>
             <enabled>true</enabled>
+            <checksumPolicy>ignore</checksumPolicy>
           </releases>
           <snapshots>
             <enabled>true</enabled>
+            <checksumPolicy>ignore</checksumPolicy>
```
cstamas, line 46: why are we ignoring checksums?

gnodet: Some tests fail with the following exception:
```
Caused by: org.eclipse.aether.resolution.ArtifactResolutionException: Could not transfer artifact org.apache.maven.plugin-tools:maven-script-ant:pom:3.7.1-SNAPSHOT from/to local.central (file:///Users/gnodet/.m2/repository): Checksum validation failed, no checksums available
```

cstamas: Ah, well, local repository should not be ..... :smile:

maven-plugin-tools-annotations/pom.xml
```diff
@@ -151,6 +151,10 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+      </plugin>
```
slawekjaranowski, line 157: should be add by default

slawekjaranowski:
one for build log ...

```
The following dependencies are in wrong scope:
 * org.apache.maven:plexus-utils:jar:4.0.0-alpha-2:compile
```

maven-plugin-plugin/src/it/v4api/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+  <groupId>org.apache.maven.its</groupId>
+  <artifactId>v4api</artifactId>
+  <version>1.0-SNAPSHOT</version>
+  <packaging>maven-plugin</packaging>
+
+  <name>Maven Integration Test :: v4api</name>
+  <description>
+    Test maven-plugin-plugin support for the Maven 4.x api.
+  </description>
+
```
slawekjaranowski, line 34: I would like to add prerequisites to inform about required version for Maven 3

slawekjaranowski:
Trying:

```
mvn org.apache.maven.its:v4api:1.0-SNAPSHOT:first

Apache Maven 4.0.0-alpha-2 (c07700ffc97d3967a8de77d5a1065f5e6115c00e)

[ERROR] Failed to parse plugin descriptor for org.apache.maven.its:v4api:1.0-SNAPSHOT (/Users/.../.m2/repository/org/apache/maven/its/v4api/1.0-SNAPSHOT/v4api-1.0-SNAPSHOT.jar): JAR entry META-INF/maven/lifecycle.xml not found in /Users/.../.m2/repository/org/apache/maven/its/v4api/1.0-SNAPSHOT/v4api-1.0-SNAPSHOT.jar
```

```
mvn org.apache.maven.its:v4api:1.0-SNAPSHOT:first

Apache Maven 3.8.6 (84538c9988a25aec085021c365c560670ad80f63)

[ERROR] Failed to parse plugin descriptor for org.apache.maven.its:v4api:1.0-SNAPSHOT (/Users/.../.m2/repository/org/apache/maven/its/v4api/1.0-SNAPSHOT/v4api-1.0-SNAPSHOT.jar): JAR entry META-INF/maven/lifecycle.xml not found in /Users/.../.m2/repository/org/apache/maven/its/v4api/1.0-SNAPSHOT/v4api-1.0-SNAPSHOT.jar -> [Help 1]
```

gnodet:
> one for build log ...
> 
> ```
> The following dependencies are in wrong scope:
>  * org.apache.maven:plexus-utils:jar:4.0.0-alpha-2:compile
> ```

`plexus-utils` is not provided anymore.  I think the plugin is fooled by the `org.apache.maven` groupid, so I think this is correct. See https://issues.apache.org/jira/browse/MNG-6965

maven-plugin-tools-annotations/pom.xml
gnodet: Removed

gnodet:
> Trying:
> 
> ```
> mvn org.apache.maven.its:v4api:1.0-SNAPSHOT:first
> 
> Apache Maven 4.0.0-alpha-2 (c07700ffc97d3967a8de77d5a1065f5e6115c00e)
> 
> [ERROR] Failed to parse plugin descriptor for org.apache.maven.its:v4api:1.0-SNAPSHOT (/Users/.../.m2/repository/org/apache/maven/its/v4api/1.0-SNAPSHOT/v4api-1.0-SNAPSHOT.jar): JAR entry META-INF/maven/lifecycle.xml not found in /Users/.../.m2/repository/org/apache/maven/its/v4api/1.0-SNAPSHOT/v4api-1.0-SNAPSHOT.jar
> ```
> 
>  ```
> mvn org.apache.maven.its:v4api:1.0-SNAPSHOT:first
> 
> Apache Maven 3.8.6 (84538c9988a25aec085021c365c560670ad80f63)
> 
> [ERROR] Failed to parse plugin descriptor for org.apache.maven.its:v4api:1.0-SNAPSHOT (/Users/.../.m2/repository/org/apache/maven/its/v4api/1.0-SNAPSHOT/v4api-1.0-SNAPSHOT.jar): JAR entry META-INF/maven/lifecycle.xml not found in /Users/.../.m2/repository/org/apache/maven/its/v4api/1.0-SNAPSHOT/v4api-1.0-SNAPSHOT.jar -> [Help 1]
> ```

Do we need the plugin to actually run ? 
I can either remove the `@Execute` or add a `lifecycle.xml`
After that, another error happens because the plugin is being injected with a custom `@Component`, but there's no such public component in the v4 api.   I can also remove the custom component injection.  But the IT is testing the `plugin.xml` generation and not the plugin itself.
Another option is to document that this plugin does not actually work.

gnodet:
> > one for build log ...
> > ```
> > The following dependencies are in wrong scope:
> >  * org.apache.maven:plexus-utils:jar:4.0.0-alpha-2:compile
> > ```
> 
> `plexus-utils` is not provided anymore. I think the plugin is fooled by the `org.apache.maven` groupid, so I think this is correct. See https://issues.apache.org/jira/browse/MNG-6965

I've removed the warning.

slawekjaranowski:
> This looks like a bug in https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/internal/impl/PropertiesAsMap.java

Yes - iterator next method always return the same element 😄

maven-plugin-plugin/src/it/v4api/pom.xml
slawekjaranowski: as we assume that is not runnable, it is never mind

slawekjaranowski (approved):
As plugin under test is only used for generating descriptor - it is ok for me.

PR #120, opened 2022-07-14
[MPLUGIN-407] Remove duplicate code from PluginReport

slawekjaranowski:


michael-o (approved):
You should also add to the deprecations that those params will be removed.

jira-importer:
Resolve #714

PR #125, opened 2022-07-24
[MPLUGIN-412] Use Maven core artifacts in provided scope

slawekjaranowski:
in production and in IT tests


maven-plugin-plugin/src/it/annotation-with-inheritance-from-deps/pom.xml
```diff
@@ -43,31 +43,33 @@ under the License.
...
     </dependency>
     <dependency>
       <groupId>org.apache.maven.plugin-tools</groupId>
       <artifactId>maven-plugin-tools-annotations</artifactId>
       <version>@project.version@</version>
       <classifier>tests</classifier>
+      <exclusions>
+        <exclusion>
+          <groupId>*</groupId>
+          <artifactId>*</artifactId>
```
michael-o, line 68: This deserves a comment, no? At least on the commit message.

maven-plugin-plugin/src/it/mplugin-272_java8/pom.xml
```diff
@@ -32,11 +32,13 @@
...
             <version>@mavenVersion@</version>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.maven.plugin-tools</groupId>
             <artifactId>maven-plugin-annotations</artifactId>
             <version>@project.version@</version>
+            <scope>provided</scope>
         </dependency>
     </dependencies>
```
michael-o, line 43: I can see that this file uses a 4-space indent instead of 2.

michael-o (changes requested):

maven-plugin-plugin/src/it/annotation-with-inheritance-from-deps/pom.xml
slawekjaranowski: comments added

michael-o (approved):

PR #126, opened 2022-07-24
[MPLUGIN-410] Create implementation attribute for @Parameter

slawekjaranowski:


olamy (approved):

maven-plugin-annotations/src/main/java/org/apache/maven/plugins/annotations/Parameter.java
```diff
@@ -72,6 +72,13 @@
      */
     String defaultValue() default "";
 
+    /**
+     * Defines the implementation in the case the parameter type is an interface.
+     *
+     * @return the implementation class name
+     */
+    Class<?> implementation() default Object.class;
```
michael-o, line 80: Question: what if the class will only be available at runtime by the user, how can it be provided here? The previous code used strings, no?

slawekjaranowski: When user uses annotation in their code, they have access to all classes. 

In plugin-tools we operate on Type from asm, and finally on class name as string, so we needn't access to class.

I hope that define property as class will be less error-prone from user perspective.

The previous implementation was only for javadoc tags ... so must be string.

michael-o: So the rationale is that if you know that you need a specific implementation you need to have it in `compile` scope?

slawekjaranowski: The same is in https://github.com/apache/maven-plugin-tools/blob/master/maven-plugin-annotations/src/main/java/org/apache/maven/plugins/annotations/Component.java

michael-o: I don't agree here. The `role` is the interface, no? The hint is the named impl as a string. I don't think that it compares. Am I wrong?

slawekjaranowski: Ok, I think it is never mind (probably) if implementation will be string or class for `parameter` annotation is used in plugin itself not in plugin configuration.
Finally class must be on plugin classpath.
 
We also have `implementation` attribute for configuration ... https://maven.apache.org/guides/mini/guide-configuring-plugins.html#mapping-complex-objects

Now I don't know which one wins implementation from parameter annotation / javadoc tags ... or from configuration ... 

😄

michael-o: @hboutemy Do you have an opinion on this?

michael-o: Correct, tricky one... @cstamas WDYT?

slawekjaranowski: currently when user use javadoc tags can deliver such information ... of course as string without verification on code level ...

michael-o: OK, I'd still to `Class`, but like to ultimately here @cstamas on this.

michael-o (approved):

jira-importer:
Resolve #724

PR #130, opened 2022-08-01
[MPLUGIN-400] Add system requirements history section

slawekjaranowski:
- add new section in report
- move report to separate package
- not use modello for simple case


maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginReport.java
```diff
@@ -159,12 +159,42 @@
...
+     * or as xml:
+     *
+     * <pre>
+     * &lt;requirementsHistories&gt;
+     *   &lt;requirementsHistory&gt;
+     *     &lt;version&gt;plugin version&lt;/version&gt;
+     *     &lt;maven&gt;maven version&lt;/maven&gt;
+     *     &lt;jdk&gt;jdk version&lt;/jdk&gt;
+     *   &lt;/requirementsHistory&gt;
+     * &lt;/requirementsHistories&gt;
```
michael-o, line 190: I wouldn't support both, it just causes confusion and inconsistency. I'd go with the second approach.

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/report/Requirements.java
```diff
@@ -0,0 +1,91 @@
...
+        final StringBuilder sb = new StringBuilder( "Requirements{" );
+        sb.append( "maven='" ).append( maven ).append( '\'' );
+        sb.append( ", jdk='" ).append( jdk ).append( '\'' );
+        sb.append( ", memory='" ).append( memory ).append( '\'' );
+        sb.append( ", diskSpace='" ).append( diskSpace ).append( '\'' );
+        sb.append( ", others=" ).append( others );
+        sb.append( '}' );
+        return sb.toString();
+    }
+}
```
michael-o, line 91: All elements which will go away should be deprecated.

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/report/RequirementsHistory.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Plugin requirements history.
+ *
+ * @author Slawomir Jaranowski
+ */
+public class RequirementsHistory
+{
+    private static final Pattern REQUIREMENT_PATTERN =
+        Pattern.compile( "\\s*(.+?)\\s+requires\\s+Maven\\s+(.+?)\\s+and\\s+JDK\\s+(.+?)\\s*" );
```
michael-o, line 35: See previous comment.

michael-o (changes requested):

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/report/Requirements.java
```diff
@@ -0,0 +1,94 @@
...
+
+import java.util.Properties;
+
+/**
+ * Plugin requirements.
+ *
+ * @deprecated will be removed in the next major version
+ */
+@Deprecated
+public class Requirements
```
michael-o, line 30: Wait a second, not the entire class will go, but only some elements, no?

slawekjaranowski: Entire class should be deprecated as parameter `requirements` is deprecated.
When we remove parameter `requirements` class will also be removed.

michael-o: Now I understand, the histories replace it altogether.

slawekjaranowski: histories will be used manually by the users to point historical requirements. 
Current requirements should be computed based on current project configuration.

michael-o (approved):
LGTM

jira-importer:
Resolve #706

PR #146, opened 2022-10-12
Refresh download page

slawekjaranowski:
Discussed at https://github.com/apache/maven-ear-plugin/pull/78

### DIFF
--- a/src/site/xdoc/download.xml.vm
+++ b/src/site/xdoc/download.xml.vm
@@ -23,102 +23,51 @@ under the License.
   <properties>
     <title>Download ${project.name} Source</title>
   </properties>
+
   <body>
     <section name="Download ${project.name} ${project.version} Source">
 
-      <p>${project.name} ${project.version} is distributed in source format. Use a source archive if you intend to build
-      ${project.name} yourself. Otherwise, simply use the ready-made binary artifacts from central repository.</p>
+      <p><strong>${project.name} ${project.version}</strong> is distributed in source format.</p>
 
-      <p>You will be prompted for a mirror - if the file is not found on yours, please be patient, as it may take 24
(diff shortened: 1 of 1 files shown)


gnodet (approved):

gnodet:
@slawekjaranowski I think the PR missed the removal of https://github.com/apache/maven-plugin-tools/blob/master/src/site/resources/download.cgi ... ?

src/site/xdoc/download.xml.vm
```diff
@@ -23,102 +23,51 @@ under the License.
...
+              <th></th>
+              <th>Link</th>
+              <th>Checksum</th>
+              <th>Signature</th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr>
+              <td>${project.name} ${project.version} (Source zip)</td>
+              <td><a href="[preferred]maven/plugin-tools/${project.artifactId}-${project.version}-source-release.zip">${project.artifactId}-${project.version}-source-release.zip</a></td>
```
gnodet, line 54: Should n't the link be `https://downloads.apache.org/maven/plugin-tools/...` ?

slawekjaranowski: `[preferred]` will be replaced by cgi to cdn address

https://infra.apache.org/release-download-pages.html#custom

so we need cgi also

gnodet: But that was when there were mirrors, which is not the case anymore IIUC, and that's why the pages have to be updated.  Do I miss something ?

gnodet: Ok, I suppose the executable called by the cgi script has been updated to point to `dlcdn.apache.org`.  Let's keep it that way then.

PR #147, opened 2022-10-12
[MPLUGIN-423] Extract plugin report into its own plugin and deprecate the previous one

gnodet:
Extract the report plugin into its own module.
This is obviously a major incompatibility and should be scheduled for 4.0.


michael-o:
I think wr can do a bit better. Leave old partial code, print the depreciation warning and recommend the new plugin in 3.7.0. MOVE all reporting related tests. In 4.0 remove old code and only new code will remain.

gnodet:
> I think wr can do a bit better. Leave old partial code, print the depreciation warning and recommend the new plugin in 3.7.0. MOVE all reporting related tests. In 4.0 remove old code and only new code will remain.

Ah, I could not understand what you meant in the previous PR.  That makes sense now !

michael-o (changes requested):
Please also check these:
```
[INFO] --- maven-dependency-plugin:3.3.0:analyze (default-cli) @ maven-plugin-report-plugin ---
[WARNING] Used undeclared dependencies found:
[WARNING]    org.codehaus.plexus:plexus-container-default:jar:1.0-alpha-9:compile
[WARNING] Unused declared dependencies found:
[WARNING]    org.apache.maven.doxia:doxia-site-renderer:jar:1.11.1:compile
[WARNING]    org.apache.maven:maven-repository-metadata:jar:3.2.5:provided
[WARNING]    org.apache.maven:maven-artifact:jar:3.2.5:provided
[WARNING]    org.apache.maven:maven-settings:jar:3.2.5:provided
[WARNING]    org.eclipse.sisu:org.eclipse.sisu.plexus:jar:0.3.5:provided
[WARNING]    org.codehaus.plexus:plexus-velocity:jar:1.2:compile
[WARNING]    org.sonatype.plexus:plexus-build-api:jar:0.0.7:compile
[WARNING]    org.junit.jupiter:junit-jupiter:jar:5.9.1:test
[WARNING]    org.hamcrest:hamcrest:jar:2.2:test
```

maven-plugin-report-plugin/src/it/mplugin-394_report-encoding/invoker.properties
```diff
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 
-invoker.goals = clean plugin:report
+invoker.goals = -Dfile.encoding=CP1252 site
```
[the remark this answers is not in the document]
slawekjaranowski: yes `goals` is not only `goals` it is arguments for cli, I will create issue for m-invoker-p for change it

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginReport.java
```diff
@@ -0,0 +1,792 @@
...
+    @Parameter( defaultValue = "false", property = "maven.plugin.skip" )
+    private boolean skip;
+
+    /**
+     * Set this to "true" to skip generating the report.
+     *
+     * @since 2.8
+     */
+    @Parameter( defaultValue = "false", property = "maven.plugin.report.skip" )
+    private boolean skipReport;
```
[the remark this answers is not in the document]
gnodet: I suppose.  Should I flag the `skip` property as `@Deprecated` for removal in 4.0 then ? The `requirements` field is also flagged as deprecated, so both can be removed in 4.0.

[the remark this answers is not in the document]
michael-o: I would do this the other way around and use the same param name as with other plugins: `skip` with property `maven.plugin.report.skip`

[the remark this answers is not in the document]
gnodet: Will that work if I declare the following ?
```
@Parameter( property = "maven.plugin.report.skip" )
private boolean skip;

@Deprecated
@Parameter( property = "maven.plugin.skip" )
private boolean skipReport;
```

This way, both properties and plugin parameters would be preserved for compatibility and we can easily remove the deprecated one for 4.0...

gnodet:
> I would like to have a report in maven-plugin-plugin module which simply extends reports from new module. We can override `executeReport` methods with warning and call supper.
> 
> In such way we need not to duplicate code. What do you thing?
> 
> We should have a separate java package in separate module.

I suppose it would be a bit cleaner.  What I had in mind was to switch to 4.0-SNAPSHOT soon after the 3.7 release.  So not sure it's really worth the extra work, but I'll try it anyway.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginReport.java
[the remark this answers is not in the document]
michael-o: Yes, this looks good to me.

michael-o:
> > I would like to have a report in maven-plugin-plugin module which simply extends reports from new module. We can override `executeReport` methods with warning and call supper.
> > In such way we need not to duplicate code. What do you thing?
> > We should have a separate java package in separate module.
> 
> I suppose it would be a bit cleaner. What I had in mind was to switch to 4.0-SNAPSHOT soon after the 3.7 release. So not sure it's really worth the extra work, but I'll try it anyway.

Good point, we shouldn't put too much effort into it. The warning is decent. The move from one to another is trivial.

gnodet:
> > I would like to have a report in maven-plugin-plugin module which simply extends reports from new module. We can override `executeReport` methods with warning and call supper.
> > In such way we need not to duplicate code. What do you thing?
> > We should have a separate java package in separate module.
> 
> I suppose it would be a bit cleaner. What I had in mind was to switch to 4.0-SNAPSHOT soon after the 3.7 release. So not sure it's really worth the extra work, but I'll try it anyway.

I was about to try, but I realized there is problem with this approach. I think it makes sense to keep the ITs with the report.  However, the ITs actually need the `maven-plugin-plugin` to generate the plugins and then run the `site` goal.  This means a cyclic dependency, which would have to be broken by moving back the ITs to the `maven-plugin-plugin` or to a new module... so not really sure the end state would be really better.

gnodet:
> Please also check these:
> 
> ```
> [INFO] --- maven-dependency-plugin:3.3.0:analyze (default-cli) @ maven-plugin-report-plugin ---
> [WARNING] Used undeclared dependencies found:
> [WARNING]    org.codehaus.plexus:plexus-container-default:jar:1.0-alpha-9:compile
> [WARNING] Unused declared dependencies found:
> [WARNING]    org.apache.maven.doxia:doxia-site-renderer:jar:1.11.1:compile
> [WARNING]    org.apache.maven:maven-repository-metadata:jar:3.2.5:provided
> [WARNING]    org.apache.maven:maven-artifact:jar:3.2.5:provided
> [WARNING]    org.apache.maven:maven-settings:jar:3.2.5:provided
> [WARNING]    org.eclipse.sisu:org.eclipse.sisu.plexus:jar:0.3.5:provided
> [WARNING]    org.codehaus.plexus:plexus-velocity:jar:1.2:compile
> [WARNING]    org.sonatype.plexus:plexus-build-api:jar:0.0.7:compile
> [WARNING]    org.junit.jupiter:junit-jupiter:jar:5.9.1:test
> [WARNING]    org.hamcrest:hamcrest:jar:2.2:test
> ```

This now looks like:
```
[INFO] --- maven-dependency-plugin:3.3.0:analyze (default-cli) @ maven-plugin-report-plugin ---
[WARN] Unused declared dependencies found:
[WARN]    org.apache.maven:maven-artifact:jar:3.2.5:provided
[WARN]    org.eclipse.sisu:org.eclipse.sisu.plexus:jar:0.3.5:provided
```
I can't really get rid of those warnings.  If I remove the `maven-artifact` dependency, I end up with
```
[ERROR] 

Some dependencies of Maven Plugins are expected to be in provided scope.
Please make sure that dependencies listed below declared in POM
have set '<scope>provided</scope>' as well.

The following dependencies are in wrong scope:
 * org.apache.maven:maven-artifact:jar:3.2.5:compile
```
which is much worse imho.
If I remove `org.eclipse.sisu.plexus`, I end up with:
```
[INFO] --- maven-dependency-plugin:3.3.0:analyze (default-cli) @ maven-plugin-report-plugin ---
[WARN] Used undeclared dependencies found:
[WARN]    org.eclipse.sisu:org.eclipse.sisu.plexus:jar:0.3.5:provided
[WARN] Unused declared dependencies found:
[WARN]    org.apache.maven:maven-artifact:jar:3.2.5:provided
[WARN]    org.codehaus.plexus:plexus-container-default:jar:2.1.1:provided
```
...

michael-o (approved):
Added two followups. This looks very good to me know. Thank you very much @gnodet!

kwin:
Please don't let me rewrite https://github.com/apache/maven-plugin-tools/pull/139. This PR conflicts with the other. Can we wait with this effort until #139 is merged?

maven-plugin-report-plugin/pom.xml
```diff
@@ -0,0 +1,268 @@
...
+        </dependency>
+
+        <!-- plexus -->
+        <dependency>
+            <groupId>org.codehaus.plexus</groupId>
+            <artifactId>plexus-utils</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.codehaus.plexus</groupId>
+            <artifactId>plexus-container-default</artifactId>
```
kwin, line 109: for what do we need that apart from testing?

maven-plugin-report-plugin/src/site/apt/examples/generate-descriptor.apt.vm
```diff
@@ -0,0 +1,118 @@
+ ------
+ Configuring Generation of Plugin Descriptor
```
kwin, line 2: this still belongs exclusively to `maven-plugin-plugin`

maven-plugin-report-plugin/src/site/apt/examples/generate-help.apt.vm
```diff
@@ -0,0 +1,58 @@
+ ------
+ Configuring Generation of Help Mojo
```
kwin, line 2: not related to report

maven-plugin-report-plugin/src/site/apt/examples/using-annotations.apt.vm
```diff
@@ -0,0 +1,117 @@
+ ------
+ Using Plugin Tools Java5 Annotations
```
kwin, line 2: should only be documented with m-p-p

maven-plugin-report-plugin/src/site/apt/index.apt
```diff
@@ -0,0 +1,85 @@
...
+~~ software distributed under the License is distributed on an
+~~ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+~~ KIND, either express or implied.  See the License for the
+~~ specific language governing permissions and limitations
+~~ under the License.
+
+~~ NOTE: For help with the syntax of this file, see:
+~~ http://maven.apache.org/doxia/references/apt-format.html
+
+Maven Plugin Plugin
```
kwin, line 29: wrong index(copy & paste from m-p-p)

maven-plugin-report-plugin/src/site/apt/examples/generate-report.apt.vm
```diff
@@ -0,0 +1,50 @@
...
+ To configure the generation of the documentation reports, add the following to the project's POM:
+
++-----+
+<project>
+  ...
+  <reporting>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-plugin-plugin</artifactId>
```
kwin, line 40: outdated

maven-plugin-report-plugin/src/site/apt/usage.apt
```diff
@@ -0,0 +1,48 @@
...
+  The Plugin Report plugin is generally used for Maven plugins. Its goals are bound to different phases in the build life cycle.
+  So when you execute a specific phase, the Plugin Plugin goals bound to it are also executed. Aside from this, each goal can
+  also be explicitly executed from the command line.
+
+* The <<<plugin-report:report>>> Goal
+
+  To generate the documentation of your plugin, execute the following on the command line:
+
++-----+
+mvn plugin:report
```
kwin, line 41: rather `mvn plugin-report:report`

kwin (changes requested):
Ok, overall, but a lot of copy&paste in the site of m-plugin-report-p

gnodet:
@kwin I've fixed the dependencies and web site.

jira-importer:
Resolve #732

PR #152, opened 2022-10-16
[MPLUGIN-426] add @Description annotation

bmarwell:
ITs missing, tests missing, draft.

### DIFF
--- a/maven-plugin-annotations/src/main/java/org/apache/maven/plugins/annotations/Description.java
+++ b/maven-plugin-annotations/src/main/java/org/apache/maven/plugins/annotations/Description.java
@@ -45,7 +45,7 @@ public @interface Description
      *
      * @return a description of the Mojo or the parameter.
      */
-    String content();
+    String value();
 
     /**
      * The version of the plugin since when this goal or parameter was introduced (inclusive, optional).
--- a/maven-plugin-annotations/src/main/java/org/apache/maven/plugins/annotations/Description.java
+++ b/maven-plugin-annotations/src/main/java/org/apache/maven/plugins/annotations/Description.java
@@ -40,7 +40,7 @@ public @interface Description
      * Description content for the {@code Mojo} or Mojo {@code Parameter}.
      *
      * <p>A &quot;Safe HTML&quot; subset can be used. This is achieved by running
-     * the content through the <a href="https://github.com/owasp/java-html-sanitizer"<OWASP Java HTML Sanitizer</a>
+     * the content through the <a href="https://github.com/owasp/java-html-sanitizer">OWASP Java HTML Sanitizer</a>
      * before rendering.</p>
      *
      * @return a description of the Mojo or the parameter.
--- a/maven-plugin-annotations/src/main/java/org/apache/maven/plugins/annotations/Description.java
+++ b/maven-plugin-annotations/src/main/java/org/apache/maven/plugins/annotations/Description.java
@@ -55,25 +55,4 @@ public @interface Description
      */
     String since() default "";
 
-    /**
-     * Marks this Mojo goal or parameter as deprecated.
-     *
-     * @return {@code true} whether this Mojo goal or parameter is deprecated.
-     */
-    boolean deprecated() default false;
-
-    /**
-     * Deprecation reason (optional).
-     *
-     * @return an empty String if no deprecation reason has been given or a description.
-     */
-    String deprecatedBecause() default "";
-
-    /**
-     * Version since when this goal or parameter has been deprecated.
-     *
-     * @return the version since when this goal or parameter has been deprecated (inclusive) or an empty String if no
-     * version was given.
-     */
-    String deprecatedSince() default "";
 }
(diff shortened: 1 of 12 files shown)


maven-plugin-annotations/src/main/java/org/apache/maven/plugins/annotations/Description.java
```diff
@@ -0,0 +1,79 @@
...
+     */
+    String deprecatedBecause() default "";
+
+    /**
+     * Version since when this goal or parameter has been deprecated.
+     *
+     * @return the version since when this goal or parameter has been deprecated (inclusive) or an empty String if no
+     * version was given.
+     */
+    String deprecatedSince() default "";
```
bmarwell, line 78: Unused atm, wdyt? was in original proposal

maven-plugin-annotations/src/main/java/org/apache/maven/plugins/annotations/Description.java
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * Description content for the {@code Mojo} or Mojo {@code Parameter}.
+     *
+     * <p>A &quot;Safe HTML&quot; subset can be used. This is achieved by running
+     * the content through the <a href="https://github.com/owasp/java-html-sanitizer"<OWASP Java HTML Sanitizer</a>
+     * before rendering.</p>
+     *
+     * @return a description of the Mojo or the parameter.
+     */
+    String content();
```
bmarwell, line 48: Naming?

maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/JavaAnnotationsMojoDescriptorExtractor.java
```diff
@@ -288,16 +288,19 @@ protected void populateDataFromJavadoc( Map<String, MojoAnnotatedClass> mojoAnno
...
-                if ( since != null )
+                if ( since != null && StringUtils.isEmpty( mojoAnnotationContent.getSince() ) )
                 {
                     mojoAnnotationContent.setSince( since.getValue() );
                 }
 
                 DocletTag deprecated = findInClassHierarchy( javaClass, "deprecated" );
-                if ( deprecated != null )
+                if ( deprecated != null && StringUtils.isEmpty( mojoAnnotationContent.getDeprecated() ) )
                 {
```
bmarwell, line 304: New behaviour (compatible): Only set from javadoc annotation if not already set by Java annotation..

maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/scanner/DefaultMojoAnnotationsScanner.java
```diff
@@ -306,11 +318,21 @@ protected void analyzeVisitors( MojoClassVisitor mojoClassVisitor )
 
                 populateAnnotationContent( parameterAnnotationContent, fieldAnnotationVisitor );
 
+                DescriptionAnnotationContent descriptionAnnotationContent = new DescriptionAnnotationContent();
+                final MojoAnnotationVisitor descriptionAnnotationVisitor =
+                        annotationVisitorMap.get( Description.class.getName() );
+                if ( descriptionAnnotationVisitor != null )
+                {
+                    populateAnnotationContent( descriptionAnnotationContent, descriptionAnnotationVisitor );
+                }
```
bmarwell, line 327: `@Description` is optional and thus can be `null`.

maven-plugin-annotations/src/main/java/org/apache/maven/plugins/annotations/Description.java
pzygielo: With this named as `value`, could annotation be used as `@Description("xxx yyy")`?

bmarwell: Yes, that was my first idea. But since it contains HTML, I thought it might be a good idea to not use "value". But then, what would value be if added later...
Can change that. Will also add tests.

cstamas:
Looks ok to me

hboutemy:
commits added as a review
build is failing on GH: does not seem related to this updates, but we'll need to fix because additional failure (like the rat failure I fixed) are not visible

gnodet:
Should the new annotation contain a `since` field to be able to avoid the `since` javadoc tag ?

bmarwell:
> Should the new annotation contain a `since` field to be able to avoid the `since` javadoc tag ?

Sounds like a plan!
But we still have a "dispute", some favour an approach where we instead implement a kotlin kdoc, scala doc and groovy doc parser as well. I am pretty sure if we went that way (ie support external parsers for other JVM languages), we should provide the appropriate parsers/extractors as well. This option here is WAY simpler, but not necessarily better.
Maybe this should go to the mailing list AGAIN? WDYT?

jira-importer:
Resolve #740

slachiewicz:
Closing: the branch has conflicted with `master` since 2022 and the design question (annotation versus doc-comment parsers) was never settled. #740 stays open as the tracker for the idea.

PR #156, opened 2022-10-24
[MPLUGIN-419] Fix documentation for maven-plugin-report-plugin

slawekjaranowski:

PR #157, opened 2022-10-25
[MPLUGIN-432] Improve site build

slawekjaranowski:
- rename duplicate packages - cause NPE in javadoc on clean project
- remove reports defined in parent pom
- aggregate reports should be not inherited


gnodet (approved):

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/DescriptorGeneratorMojo.java
```diff
@@ -195,7 +194,7 @@
      * This may be relative to the root of the generated Maven site.
      * It does not need to exist yet at the time when this goal is executed.
      * Must end with a slash.
-     * <b>In case this is set the javadoc reporting goal should be executed prior to {@link PluginReport}.</b>
+     * <b>In case this is set the javadoc reporting goal should be executed prior to Plugin Report.</b>
```
kwin, line 197: I would still put a javadoc reference there (to the new goal). This is linked to the javadoc page in the goal page.

kwin (approved):
Looks good apart from the minor comment regarding the javadoc link

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/DescriptorGeneratorMojo.java
slawekjaranowski: I will do it with #156

jira-importer:
Resolve #733

PR #160, opened 2022-10-26
[MPLUGIN-431] Remove deprecated items from new maven-plugin-report-plugin

slawekjaranowski:


maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/AbstractGeneratorMojo.java
```diff
@@ -107,7 +107,11 @@ else if ( !goalPrefix.equals( defaultGoalPrefix ) )
     static String getDefaultGoalPrefix( MavenProject project )
     {
         String defaultGoalPrefix;
-        if ( "maven-plugin".equalsIgnoreCase( project.getArtifactId() ) )
+        if ( "maven-plugin-report-plugin".equalsIgnoreCase( project.getArtifactId() ) )
+        {
+            defaultGoalPrefix = "plugin-report";
```
gnodet, line 112: Is that needed ?
I forced at https://github.com/apache/maven-plugin-tools/blob/master/maven-plugin-report-plugin/pom.xml#L182

slawekjaranowski: we have warning during build

```
[WARNING] 

Goal prefix is specified as: 'plugin-report'. Maven currently expects it to be 'report'.
```

so I want to mitigate it for next release

gnodet: Won't that be the case for all plugins if we start splitting the mojos from the reports ?

slawekjaranowski: No - by default `maven` and `plugin` words are removed from artifactId, so it is special case.

gnodet: Changed my mind. LGTM

gnodet (approved):

jira-importer:
Resolve #738

PR #162, opened 2022-10-26
[MPLUGIN-433] Allow to disable link validation

kwin:
Also validate internal links to javadocs in mojo/parameter description and deprecated

### DIFF
--- a/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/report_old/PluginReport.java
+++ b/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/report_old/PluginReport.java
@@ -219,6 +219,19 @@ public class PluginReport
     private File enhancedPluginXmlFile;
 
     /**
+     * In case the internal javadoc site has not been generated when running this report goal
+     * (e.g. when using an aggregator javadoc report) link validation needs to be disabled by setting
+     * this value to {@code true}.
+     * This might have the drawback that some links being generated in the report might be broken
+     * in case not all parameter types and javadoc link references are resolvable through the sites being given to
+     * {@link DescriptorGeneratorMojo}.
+     * 
+     * @since 3.7.0
+     */
+    @Parameter
+    private boolean disableInternalJavadocLinkValidation;
+
+    /**
      * {@inheritDoc}
      */
     @Override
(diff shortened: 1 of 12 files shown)


maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/report_old/PluginReport.java
```diff
@@ -218,6 +218,19 @@
...
+     * (e.g. when using an aggregator javadoc report) link validation needs to be disabled by setting
+     * this value to {@code true}.
+     * This might have the drawback that some links being generated in the report might be broken
+     * in case not all parameter types and javadoc link references are resolvable through the sites being given to
+     * {@link DescriptorGeneratorMojo}.
+     * 
+     * @since 3.7.0
+     */
+    @Parameter
+    private boolean disableInternalJavadocLinkValidation;
```
slawekjaranowski, line 232: I would like not to add new parameters to deprecated report - what do you think?

kwin: A lot of parameters have been added in 3.7.0 (currently both in deprecated and new report goal) not only this one. I don't think that this does any harm here.

slawekjaranowski: ok, no problem here

slawekjaranowski:
Please also rebase with current codebase ... build should will be ok 😄

kwin:
> Please also rebase with current codebase

Done.

slawekjaranowski:
What do you think about adding IT test for such options?

By the way we don't have any IT tests for javadoc links ... or I don't see.

kwin:
> By the way we don't have any IT tests for javadoc links ... or I don't see.

Not end2end but there is a lot of unit tests and https://github.com/apache/maven-plugin-tools/blob/master/maven-plugin-tools-api/src/test/java/org/apache/maven/tools/plugin/javadoc/JavadocSiteIT.java which checks that the links being created actually work against a Javadoc site created from JDK8/11/17.

slawekjaranowski (approved):

kwin:
I now also added a proper IT in https://github.com/apache/maven-plugin-tools/pull/162/commits/6b564c8aaf89af4776ffb7898e2a86a4ce400a4f with some basic checks on generated links.

jira-importer:
Resolve #731

PR #165, opened 2022-10-27
Use latest m-p-p and m-p-report-p also for the site of m-plugin-tools

kwin:


kwin:
@gnodet Any idea how to make this work?

gnodet:
> @gnodet Any idea how to make this work?

I have this exception:
```
INFO] 
[INFO] --- maven-plugin-plugin:3.7.1-SNAPSHOT:helpmojo (generated-helpmojo) @ maven-plugin-plugin ---
[INFO] Downloading from apache.snapshots: https://repository.apache.org/snapshots/org/apache/maven/plugins/maven-plugin-plugin/3.7.1-SNAPSHOT/maven-plugin-plugin-3.7.1-20221026.201321-13.pom
[INFO] Downloaded from apache.snapshots: https://repository.apache.org/snapshots/org/apache/maven/plugins/maven-plugin-plugin/3.7.1-SNAPSHOT/maven-plugin-plugin-3.7.1-20221026.201321-13.pom (11 kB at 72 kB/s)
[WARN] The POM for org.apache.maven.plugins:maven-plugin-plugin:jar:3.7.1-20221026.201321-13 is invalid, transitive dependencies (if any) will not be available, enable verbose output (-X) for more details
[WARN] Error injecting: org.apache.maven.plugin.plugin.HelpGeneratorMojo
java.lang.NoClassDefFoundError: org/apache/maven/tools/plugin/generator/GeneratorException
	at java.base/java.lang.Class.getDeclaredConstructors0(Native Method)
```
I don't think we can have the plugin being packaged with itself.  Maven has to load the jar, but it has not been built yet.

pom.xml
```diff
@@ -325,6 +325,22 @@
...
+            <externalJavadocBaseUrls>
+              <externalJavadocBaseUrl>https://docs.oracle.com/javase/8/docs/api/</externalJavadocBaseUrl>
+            </externalJavadocBaseUrls>
+          </configuration>
+        </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-plugin-report-plugin</artifactId>
+          <version>${project.version}</version>
+        </plugin>
```
slawekjaranowski, line 343: IMHO we can't use it now in this way ... there is egg and hen problem
We should use the latest released version

The `maven-plugin-report-plugin` has not been released yet

maven-plugin-plugin/pom.xml
```diff
@@ -271,7 +271,7 @@
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-plugin-plugin</artifactId>
+        <artifactId>maven-plugin-report-plugin</artifactId>
```
slawekjaranowski, line 274: Here we will have a cycle ... `maven-plugin-report-plugin` also requires `maven-plugin-plugin`

slawekjaranowski:
> So this may possible for the site which uses `m-plugin-report-p`, but not for `m-plugin-p` which uses... itself.

But we must do it in two steps .... first with `install` and second with `site` - I would like not to prefer do it in this way

kwin:
> But we must do it in two steps .... first with `install` and second with `site` - I would like not to prefer do it in this way

Even if you do that it fails. Please try it locally. But if there is no easy way to achieve that, I am fine with sticking to the previous release.

PR #167, opened 2022-10-28
[MPLUGIN-437] Fixes to the plugin descriptor generation

gnodet:
When a mojo inherits from a mojo in a different project, the plugin descriptor generation fails.
Another problem is when the source artifact is being built in the same reactor, I had a case where the resolved artifacts file was pointing to a directory.

### DIFF
--- a/maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/converter/JavaClassConverterContext.java
+++ b/maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/converter/JavaClassConverterContext.java
@@ -34,7 +34,6 @@ import java.util.stream.Collectors;
 import com.thoughtworks.qdox.JavaProjectBuilder;
 import com.thoughtworks.qdox.builder.TypeAssembler;
 import com.thoughtworks.qdox.library.ClassNameLibrary;
-import com.thoughtworks.qdox.library.SourceLibrary;
 import com.thoughtworks.qdox.model.JavaClass;
 import com.thoughtworks.qdox.model.JavaField;
 import com.thoughtworks.qdox.model.JavaModule;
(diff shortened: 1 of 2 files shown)


slawekjaranowski:
Can you create issue for it?

slawekjaranowski (approved):

maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/converter/JavaClassConverterContext.java
```diff
@@ -81,13 +80,6 @@ public JavaClassConverterContext( JavaClass mojoClass, JavaClass declaringClass,
                                       Map<String, MojoAnnotatedClass> mojoAnnotatedClasses,
                                       JavadocLinkGenerator linkGenerator, int lineNumber )
     {
-        if ( !( mojoClass.getJavaClassLibrary() instanceof SourceLibrary )
```
kwin, line 84: why has this been removed? Access to import statements is still crucial for resolving javadoc references.

kwin (changes requested):

maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/converter/JavaClassConverterContext.java
gnodet: I did remove the fact that the generation will throw an exception, not the fact that it will be used.

kwin: I understand this, but how can it work correctly? The exception was a good indicator to ease debugging.

gnodet: Well, in my case, the parent class was not part of the module being built, and the `declaringClass.getJavaClassLibrary()` is an instance of `ClassLoaderLibrary`.
The generation will still work, even if some links can not be resolved.  But the plugin as a whole can not fail for this reason.  And it's a valid use case, so even a warning may be too much, unless it can be disabled.

kwin: I think at least a warning is reasonable if javadoc links cannot be resolved.

gnodet: > Also how can meta information from the parent mojos be extracted which are not part of the binary (i.e. coming from javadoc like description, since or deprecated)? I think this requires evaluation of the parent's plugin descriptor....

I guess they don't.  
Instead of warning upfront, I think it would be better to display a warning at the time the information can not be retrieved.  That would avoid irrelevant warnings to be displayed.  For example if the parent class does not contain any parameter definition, a warning would be irrelevant noise.

jira-importer:
Resolve #744

PR #168, opened 2022-10-29
[MPLUGIN-438] Parameter description should be taken from annotated item

slawekjaranowski:


maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/JavaAnnotationsMojoDescriptorExtractor.java
```diff
@@ -463,6 +472,51 @@ private Map<String, JavaAnnotatedElement> extractParameterAnnotations( JavaClass
...
+     * @return map with Mojo parameters names as keys
+     */
+    private Map<String, JavaAnnotatedElement> extractMethodsAnnotations( JavaClass javaClass,
+                                                                        Map<String, JavaClass> javaClassesMap )
+    {
+        try
+        {
+            Map<String, JavaAnnotatedElement> rawParams = new TreeMap<>();
+
+            // we have to add the parent fields first, so that they will be overwritten by the local fields if
```
kwin, line 497: methods instead of fields

maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/datamodel/ParameterAnnotationContent.java
```diff
@@ -48,20 +48,24 @@
...
         super( fieldName );
         this.className = className;
         this.typeParameters = typeParameters;
+        this.methodSource = methodSource;
     }
 
     public ParameterAnnotationContent( String fieldName, String alias, String property, String defaultValue,
                                        boolean required, boolean readonly, String className,
-                                       List<String> typeParameters )
+                                       List<String> typeParameters, boolean methodSource )
```
kwin, line 66: maybe `annotationOnMethod` or `methodAnnotation`?

kwin (approved):
Overall looks fine, just some minor nitpicks. Maybe an IT would be nice as well.

maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/datamodel/ParameterAnnotationContent.java
slawekjaranowski: change to `annotationOnMethod`

PR #170, opened 2022-11-01
[MPLUGIN-439] Support custom phases in Execute annotation

kwin:
Add Unit test for JavaAnnotationsMojoDescriptorExtractor


maven-plugin-annotations/src/main/java/org/apache/maven/plugins/annotations/Execute.java
```diff
@@ -39,20 +39,31 @@
...
 
     /**
-     * goal to fork. Note that specifying a phase overrides specifying a goal. The specified <code>goal</code> must be
+     * Custom lifecycle phase to fork. Note that specifying a {@link #phase} overrides the specified custom phase.
+     * Also specifying a custom phase overrides specifying a goal.
+     * This element should only be used for non-standard phases. For standard phases rather use {@link #phase()}.
+     * 
+     * @since 3.8.0
+     */
+    String customPhase() default "";
```
slawekjaranowski, line 56: maybe an error or at least warning if user defined both
In many case users don't read a documentation.

kwin: Agree, with https://github.com/apache/maven-plugin-tools/pull/170/commits/9ccc3223cf3e48e01668c72e15d58b9b05a25577 an exception is thrown in that case.

slawekjaranowski (approved):

gnodet:
@kwin do you plan to create a similar PR for api/maven-api-core/src/main/java/org/apache/maven/api/plugin/annotations/Execute.java ?

kwin:
@gnodet Done in https://github.com/apache/maven/pull/865.

PR #171, opened 2022-11-04
[MPLUGIN-425] Add required Java and Maven version to the generated

kwin:
plugin descriptor

### DIFF
--- a/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/DescriptorGeneratorMojo.java
+++ b/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/DescriptorGeneratorMojo.java
@@ -251,6 +251,7 @@ public class DescriptorGeneratorMojo
      * compilation (determined by the extractor).</li>
      * <li>{@code "none"} to not set a required Java version.</li>
      * </ul>
+     * 
      * @since 3.8.0
      */
     @Parameter( defaultValue = VALUE_AUTO )
@@ -265,7 +266,7 @@ public class DescriptorGeneratorMojo
      * syntax like {@code "[2.0.10,2.1.0),[3.0,)"} or use a single version like {@code "2.2.1"}. The latter is a short
      * form for {@code "[2.2.1,)"}, i.e. denotes the minimum version required.</li>
      * <li>{@code "auto"} to determine the minimum Maven version from the POMs Maven prerequisite, or if not set the
-     * referenced Maven API version.</li>
+     * referenced Maven Plugin API version.</li>
      * <li>{@code "none"} to not set a required Maven version.</li>
      * </ul>
      * This value (if not set to {@code "none"}) takes precedence over the 
--- a/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/DescriptorGeneratorMojo.java
+++ b/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/DescriptorGeneratorMojo.java
@@ -41,6 +41,7 @@ import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.plugins.annotations.ResolutionScope;
 import org.apache.maven.settings.Settings;
 import org.apache.maven.tools.plugin.DefaultPluginToolsRequest;
+import org.apache.maven.tools.plugin.ExtendedPluginDescriptor;
 import org.apache.maven.tools.plugin.PluginToolsRequest;
 import org.apache.maven.tools.plugin.extractor.ExtractionException;
 import org.apache.maven.tools.plugin.generator.GeneratorException;
@@ -70,6 +71,9 @@ import org.sonatype.plexus.build.incremental.BuildContext;
 public class DescriptorGeneratorMojo
     extends AbstractGeneratorMojo
 {
+    private static final String VALUE_AUTO = "auto";
+    private static final String VALUE_NONE = "none";
+
     /**
      * The directory where the generated <code>plugin.xml</code> file will be put.
      */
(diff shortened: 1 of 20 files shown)


maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/DescriptorGeneratorMojo.java
```diff
@@ -235,6 +239,44 @@
...
+     * The required Maven version to set in the plugin descriptor. This is evaluated by Maven 4 and ignored by earlier
+     * Maven versions. Can be either one of the following formats:
+     * 
+     * <ul>
+     * <li>A version range which specifies the supported Maven versions. It can either use the usual mathematical
+     * syntax like {@code "[2.0.10,2.1.0),[3.0,)"} or use a single version like {@code "2.2.1"}. The latter is a short
+     * form for {@code "[2.2.1,)"}, i.e. denotes the minimum version required.</li>
+     * <li>{@code "auto"} to determine the minimum Maven version from the POMs Maven prerequisite, or if not set the
+     * referenced Maven Plugin API version.</li>
+     * <li>{@code "none"} to not set a required Maven version.</li>
```
slawekjaranowski, line 270: I would like to require always java version and maven version
It will resolve MPLUGIN-418

maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/JavaAnnotationsMojoDescriptorExtractor.java
```diff
@@ -102,6 +106,36 @@
 
     private static final GroupKey GROUP_KEY = new GroupKey( GroupKey.JAVA_GROUP, 100 );
 
+    /**
+     * 
+     * @see <a href="https://docs.oracle.com/javase/specs/jvms/se19/html/jvms-4.html#jvms-4.1">JVMS 4.1</a>
+     */
+    private static final Map<Integer, String> CLASS_VERSION_TO_JAVA_STRING;
```
slawekjaranowski, line 113: We should add reference to this code from pom to remember that we need update together with asm

maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/scanner/DefaultMojoAnnotationsScanner.java
```diff
@@ -85,12 +87,18 @@ public Map<String, MojoAnnotatedClass> scan( MojoAnnotationsScannerRequest reque
             for ( Artifact dependency : request.getDependencies() )
             {
                 scan( mojoAnnotatedClasses, dependency.getFile(), request.getIncludePatterns(), dependency, true );
+                if ( request.getMavenApiVersion() == null
+                     && dependency.getFile().getName().contains( "maven-plugin-api" ) )
+                {
+                    request.setMavenApiVersion( getSpecificationVersionOfJar( dependency.getFile() ) );
+                }
```
slawekjaranowski, line 94: Why not detect on Mojo level from dependencies?

kwin: I don’t understand. This value is actually derived from the maven-plugin-api dependency. Per plugin there is only one classpath and dependency so this cannot differ per mojo…

slawekjaranowski: ok, other question

why not use 

```
                dependency.getArtifactId();
                dependency.getVersion();
```

instead of checking artifact file name and next read content from artifact manifest.

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/DescriptorGeneratorMojo.java
kwin: IMHO MPLUGIN-418 is about the prerequisites in the pom.xml (relevant for Maven 3), while this PR is only about plugin descriptor (only relevant for Maven 4). Therefore even removing the "none" value here does not require a Maven prerequisite in the pom.xml. Still removing the none is probably a good idea, fixed in https://github.com/apache/maven-plugin-tools/pull/171/commits/4c8e2499919b75c2e1cd5a12e474dc65070f7d4f

maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/JavaAnnotationsMojoDescriptorExtractor.java
kwin: Done in https://github.com/apache/maven-plugin-tools/pull/171/commits/4c8e2499919b75c2e1cd5a12e474dc65070f7d4f.

maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/scanner/DefaultMojoAnnotationsScanner.java
kwin: Yes, a lot simpler, done in https://github.com/apache/maven-plugin-tools/pull/171/commits/4c8e2499919b75c2e1cd5a12e474dc65070f7d4f.

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/DescriptorGeneratorMojo.java
```diff
@@ -235,6 +238,42 @@
...
+     * Maven versions. Can be either one of the following formats:
+     * 
+     * <ul>
+     * <li>A version range which specifies the supported Maven versions. It can either use the usual mathematical
+     * syntax like {@code "[2.0.10,2.1.0),[3.0,)"} or use a single version like {@code "2.2.1"}. The latter is a short
+     * form for {@code "[2.2.1,)"}, i.e. denotes the minimum version required.</li>
+     * <li>{@code "auto"} to determine the minimum Maven version from the POMs Maven prerequisite, or if not set the
+     * referenced Maven Plugin API version.</li>
+     * </ul>
+     * This value (if not set to {@code "none"}) takes precedence over the
```
slawekjaranowski, line 269: `none` is not present

slawekjaranowski (approved):

cstamas (approved):

PR #177, opened 2022-11-20
[MPLUGIN-443] Detect javadoc link labels with spaces

kwin:


michael-o:
I am about to puke. Does javadoc really produces these?
This violates https://stackoverflow.com/a/2849800/696632 and https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

kwin:
I think there is a misunderstanding, the label will never affect the id/name of an HTML element but only the label, i.e. element value of `<a href="....">label</a>` (therefore its name :-))

maven-plugin-tools-api/src/test/java/org/apache/maven/tools/plugin/javadoc/JavadocReferenceTest.java
```diff
@@ -45,6 +45,8 @@ void testParse()
                       JavadocReference.parse( "package.Class#member(ArgType1,ArgType2) label" ) );
         assertEquals( new JavadocReference( Optional.of("my.module"), Optional.of( "package.Class" ), Optional.of( "member(ArgType1,ArgType2)" ), Optional.of("label") ),
                       JavadocReference.parse( "my.module/package.Class#member(ArgType1,ArgType2) label" ) );
+        assertEquals( new JavadocReference( Optional.empty(), Optional.empty(), Optional.of( "member" ), Optional.of("label with spaces") ),
+                JavadocReference.parse( "#member label with spaces" ) );
```
michael-o, line 49: My bad, but this id/anchor contains spaces...?!

kwin: This is no anchor. It is a Javadoc code reference.

PR #214, opened 2023-05-09
[MNG-6829] Replace any StringUtils#isEmpty(String) and #isNotEmpty(String)

timtebeek:
Continuation of https://issues.apache.org/jira/browse/MNG-6825

Use this link to re-run the recipe: https://public.moderne.io/recipes/org.openrewrite.java.migrate.apache.commons.lang.IsNotEmptyToJdk?organizationId=QXBhY2hlIE1hdmVu

Review requested of @elharo

### DIFF
--- a/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/report_old/PluginReport.java
+++ b/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/report_old/PluginReport.java
@@ -378,7 +378,7 @@ public class PluginReport extends AbstractMavenReport {
 
             this.project = project;
 
-            this.requirements = (requirements == null ? new Requirements() : requirements);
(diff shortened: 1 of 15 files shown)


elharo (approved):

michael-o:
Buit not gone from the POM, no?

timtebeek:
> Buit not gone from the POM, no?

We're not at the stage yet where we have removed all methods of StringUtils, so we can't yet remove the dependency. See for instance https://github.com/apache/maven-plugin-tools/blob/73791dd14fab1f974bab6415b074835945336728/maven-script/maven-script-ant/src/main/java/org/apache/maven/script/ant/AntMojoWrapper.java#L209-L210

As requested I'm doing small individual (automated) changes, such that they are easier to review and merge, with less risk.
We hope to get to fully removing StringUtils soon, but feel free to help out either in an automated fashion, or by cleaning up what remains after the first few rounds.

PR #215, opened 2023-05-09
Fixes

michael-o:
One PR for 3.9.0. Cleanly separated commits.


pom.xml
```diff
@@ -182,32 +180,11 @@
...
-      <!-- doxia -->
+      <!-- shared -->
       <dependency>
-        <groupId>org.apache.maven.doxia</groupId>
-        <artifactId>doxia-sink-api</artifactId>
-        <version>${doxiaVersion}</version>
-        <exclusions>
-          <exclusion>
-            <groupId>org.codehaus.plexus</groupId>
-            <artifactId>plexus-container-default</artifactId>
```
slawekjaranowski, line 193: I we have it on dependency list will generate warning in 3.9.2

maven-plugin-report-plugin/pom.xml
```diff
@@ -88,6 +86,7 @@
     <dependency>
       <groupId>org.apache.maven.doxia</groupId>
       <artifactId>doxia-sink-api</artifactId>
+      <version>1.11.1</version>
```
slawekjaranowski, line 89: Most of dependency is managed, this one will not - only small inconsistency

maven-plugin-report-plugin/src/it/fix-maven-since-3.x/pom.xml
```diff
@@ -81,7 +81,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-help-plugin</artifactId>
-        <version>2.1.1</version>
+        <version>3.4.0</version>
```
slawekjaranowski, line 84: property?

maven-plugin-report-plugin/src/it/plugin-report/verify.groovy
```diff
@@ -28,3 +28,11 @@ assert !pluginInfo.text.contains('Disk Space')
...
 assert pluginInfo.text.contains('1.8')
 assert pluginInfo.text.contains('3.2.5')
+
+assert new File( basedir, 'target/site/de/noop-mojo.html' ).isFile()
+assert new File( basedir, 'target/site/de/report-mojo.html' ).isFile()
+assert new File( basedir, 'target/site/de/plugin-info.html' ).isFile()
+
+assert new File( basedir, 'target/site/fr/noop-mojo.html' ).isFile()
+assert new File( basedir, 'target/site/fr/report-mojo.html' ).isFile()
+assert new File( basedir, 'target/site/fr/plugin-info.html' ).isFile()
```
slawekjaranowski, line 38: Should we also check `en` locale and default?

maven-plugin-report-plugin/pom.xml
michael-o: It is just a single one. I didn not create property on purpose.

pom.xml
michael-o: Which exactly? This has been deleted. You mean maven-reporting-api?

maven-plugin-report-plugin/src/it/fix-maven-since-3.x/pom.xml
michael-o: This is single use, here too, on purpose none

maven-plugin-report-plugin/src/it/plugin-report/verify.groovy
michael-o: This is already done in the script: https://github.com/apache/maven-plugin-tools/pull/215/files/4b9556f99f5335cf0f4ebba3e151f8c0157ea7d8#diff-e6a05d5a0be5ba47cd4a603f6422fb5f8467085d97800b02bc82b60f9582853dR23.

pom.xml
michael-o: This problem will be solved when we move to Doxia 2.0.0. For now, we need to exclude otherwise things start to fail. This is OK for me for now.

maven-plugin-report-plugin/pom.xml
```diff
@@ -101,13 +100,30 @@
...
     <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter-api</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <scope>test</scope>
+    </dependency>
```
slawekjaranowski, line 126: Why you add this dependencies?

pom.xml
slawekjaranowski: ok

maven-plugin-report-plugin/pom.xml
michael-o: Because otherwise the test fails for me. At least, when added all tests work.

michael-o: Remove and see whether it works for you.

michael-o: I am happy to reduce as many dependencies as possible.

michael-o: Are we good here or should I pursue another reduce? Before Doxia 2.0.0 the effort isn't really worth it.

slawekjaranowski: looks as works after remove for me - I can try clean it later.

slawekjaranowski (approved):

PR #216, opened 2023-05-23
[MPLUGIN-473] Only check direct dependencies for provided scope

gnodet:
This relaxes the check a bit and provides support for plexus-xml 4.0.0
JIRA issue: https://issues.apache.org/jira/browse/MPLUGIN-473

### DIFF
--- a/maven-plugin-plugin/src/it/mplugin-370-maven-deps-scope-bad/verify.groovy
+++ b/maven-plugin-plugin/src/it/mplugin-370-maven-deps-scope-bad/verify.groovy
@@ -19,4 +19,4 @@
 
 File buildLog = new File( basedir, "build.log" );
 assert buildLog.isFile()
-assert buildLog.text.contains( "Some dependencies of Maven Plugins are expected to be in provided scope." )
+assert buildLog.text.contains( "Some direct dependencies of Maven Plugins are expected to be in provided scope." )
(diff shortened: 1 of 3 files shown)


gnodet:
This should provide a fix for https://github.com/codehaus-plexus/plexus-utils/issues/256 and https://github.com/codehaus-plexus/plexus-xml/issues/16

slawekjaranowski:
I'm not sure if it is good way to resolve problems with plexus-utils and maven-xml* artifacts. 
Currently we have a validators in Maven core which check also transitive dependencies. 
So here verifications should be even more strict than in Maven core.
Maybe we can exclude maven-xml* artifacts?

@cstamas what do you think?

gnodet:
> I'm not sure if it is good way to resolve problems with plexus-utils and maven-xml* artifacts. Currently we have a validators in Maven core which check also transitive dependencies. So here verifications should be even more strict than in Maven core. Maybe we can exclude maven-xml* artifacts?
> 
> @cstamas what do you think?

That was a wrong idea.  The plexus-xml is actually provided by maven 3 and 4 (at least the classes), so in most cases, the `plexus-xml` dependency itself should be flagged as provided (contrary to the `plexus-utils` one).

jira-importer:
Resolve #784

PR #225, opened 2023-09-07
[MPLUGIN-442] Generate goal documentation leveraging Sink API

kwin:
Drop XDoc intermediate format


maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/GoalRenderer.java
```diff
@@ -0,0 +1,543 @@
...
+    }
+
+    private void renderParameterDetails(Iterator<Parameter> parameters) {
+
+        startSection(getI18nString("goal.parameter.details"));
+
+        while (parameters.hasNext()) {
+            Parameter parameter = parameters.next();
+            // TODO: rework parameter, currently 3 are generated: one from skin, one from startSection (all lowercase) and one manually
+            sink.anchor(parameter.getName());
```
kwin, line 326: Which anchor should we keep for backwards compatibility reasons?

maven-plugin-tools-api/src/main/java/org/apache/maven/tools/plugin/util/PluginUtils.java
```diff
@@ -137,4 +145,50 @@ public int compare(Parameter parameter1, Parameter parameter2) {
...
+     * @param project a MavenProject instance, could be null
+     * @return <code>true</code> if the Mojo class implements <code>MavenReport</code>,
+     * <code>false</code> otherwise.
+     * @throws IllegalArgumentException if any
+     * @since 3.10.0
+     */
+    public static boolean isMavenReport(String mojoClassName, MavenProject project) throws IllegalArgumentException {
+        if (mojoClassName == null) {
+            throw new IllegalArgumentException("mojo implementation should be declared");
+        }
```
michael-o, line 160: A new method should use `Objects#notNull()` here and the caller should just wrap it.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/AbstractPluginReportRenderer.java
```diff
@@ -0,0 +1,54 @@
...
+        this.locale = locale;
+        this.project = project;
+    }
+
+    /** @param key The key.
+     * @return The translated string. */
+    protected String getI18nString(String key) {
+        return i18n.getString(RESOURCE_BASENAME, locale, KEY_PREFIX + key);
+    }
+}
```
michael-o, line 54: It should follow the pattern as in here: https://github.com/apache/maven-project-info-reports-plugin/blob/ec0918223c43e3135aa067c9fcdc6c0a2427f24a/src/main/java/org/apache/maven/report/projectinfo/AbstractProjectInfoRenderer.java, maybe even with `getI18Nsection()`. This stencil is used throughout all reporting plugins now, at least in the `doxia-2.0.0` branch.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginOverviewRenderer.java
```diff
@@ -0,0 +1,392 @@
...
+
+            /*
+             * Added ./ to define a relative path
+             * @see AbstractMavenReportRenderer#getValidHref(java.lang.String)
+             */
+            String goalDocumentationLink = "./" + mojo.getGoal() + "-mojo.html";
+
+            String description;
+            if (StringUtils.isNotEmpty(mojo.getDeprecated())) {
+                description = "<strong>" + getI18nString("goal.deprecated") + "</strong> " + mojo.getDeprecated();
```
michael-o, line 129: This is a bit ugly and should rather not use then `tableCell()`, but use `Sink#tableCell()`/`Sink#tableCell()_`.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginOverviewRenderer.java
```diff
@@ -0,0 +1,392 @@
...
+            sink.tableRow();
+            tableCell(createLinkPatternedText(goalName, goalDocumentationLink));
+            if (hasMavenReport) {
+                if (PluginUtils.isMavenReport(mojo.getImplementation(), project)) {
+                    tableCell(getI18nString("isReport"));
+                } else {
+                    tableCell(getI18nString("isNotReport"));
+                }
+            }
+            tableCell(description, true);
```
michael-o, line 145: Description handling should be moved here since it appear first here.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/GoalRenderer.java
michael-o: I need to think about this and look at the output first.

michael-o (changes requested):

maven-plugin-tools-api/src/main/java/org/apache/maven/tools/plugin/util/PluginUtils.java
kwin: TBH the code was just moved from `GeneratorUtils` in maven-plugin-tools-generator in order to reduce dependencies. Let us improve the code in a separate commit/PR.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginOverviewRenderer.java
kwin: Everything in PluginOverviewRenderer has just been moved from the inner class in https://github.com/apache/maven-plugin-tools/blob/master/maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginReport.java#L262. The only refactoring was due to the newly introduced common super class (`AbstractPluginReportRenderer`). Let us track other improvements to the `PluginOverviewRenderer` separately.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginOverviewRenderer.java
kwin: see https://github.com/apache/maven-plugin-tools/pull/225#discussion_r1335221503

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginOverviewRenderer.java
```diff
@@ -0,0 +1,392 @@
...
+            sb.append("      </plugin>").append(Markup.EOL);
+            sb.append("      ...").append(Markup.EOL);
+            sb.append("    </plugins>").append(Markup.EOL);
+            sb.append("  </reporting>").append(Markup.EOL);
+        }
+
+        sb.append("  ...").append(Markup.EOL);
+        sb.append("</project>");
+
+        verbatimText(sb.toString());
```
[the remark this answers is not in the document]
kwin: https://github.com/apache/maven-plugin-tools/pull/225#discussion_r1335221503

maven-plugin-tools-api/src/main/java/org/apache/maven/tools/plugin/util/PluginUtils.java
michael-o: Agreed. Please make a note.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginOverviewRenderer.java
michael-o: Agreed. Please make a note.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginOverviewRenderer.java
michael-o: Agreed. Please make a note.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginOverviewRenderer.java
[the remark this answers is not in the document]
michael-o: Agreed. Please make a note.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/GoalRenderer.java
```diff
@@ -0,0 +1,544 @@
...
+            MavenProject project,
+            MojoDescriptor descriptor,
+            boolean disableInternalJavadocLinkValidation,
+            Log log)
+            throws IOException {
+        String filename = descriptor.getGoal() + "-mojo.html";
+        Sink sink = sinkFactory.createSink(outputDirectory, filename);
+        return new GoalRenderer(
+                sink, i18n, locale, project, descriptor, outputDirectory, disableInternalJavadocLinkValidation, log);
+    }
```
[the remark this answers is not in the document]
kwin: Fixed

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginReport.java
```diff
@@ -229,403 +195,34 @@ public String getOutputName() {
...
-                        return target.getValue();
-                    }
+        if (pluginDescriptor.getMojos() != null) {
+            for (MojoDescriptor descriptor : pluginDescriptor.getMojos()) {
+                GoalRenderer renderer;
+                try {
+                    renderer = GoalRenderer.create(
+                            getSinkFactory(),
+                            getReportOutputDirectory(),
+                            i18n,
```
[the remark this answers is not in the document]
kwin: Fixed.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginOverviewRenderer.java
```diff
@@ -0,0 +1,392 @@
...
+
+        this.pluginDescriptor = pluginDescriptor;
+
+        this.hasExtensionsToLoad = hasExtensionsToLoad;
+    }
+
+    @Override
+    public String getTitle() {
+        return getI18nString("title");
+    }
```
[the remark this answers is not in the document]
kwin: Fixed.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/AbstractPluginReportRenderer.java
kwin: Done.

kwin:
@michael-o Any more comments on the anchors being generated?

michael-o (commented):
So far so good, let me run and compare myself and think about your open question in the next couple of days...

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/GoalRenderer.java
michael-o: For as long as you don't link the anchor or need it stable (language-agnostic), there is no reason to invoke `Sink#anchor()` directly. I would drop it. The anchors from the skin are a skin problem because we don't know anything about them at this stage. So dropping 307 and 310 seems right to me.

michael-o (changes requested):

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/GoalRenderer.java
kwin: I link the anchor from the same page. Although for this use case I could use the new format created via HtmlTools.encodeId(...) called from `AbstractMavenReportRenderer.startSection(...)` potentially there exist some links leveraging those anchors externally. So we have IMHO three options:

1. Keep the anchor backwards compatible (but maybe invalid as it was done with xdoc in https://github.com/apache/maven-plugin-tools/blob/b7138b0a7744e31d43c2c16894203cd6debc1514/maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/PluginXdocGenerator.java#L649, i.e. case-sensitive names) or
2. Switch to the new format leveraging `HtmlTools.encodeId(...)` potentially breaking existing external deeplinks.
3. Create both formats old and new and use new one internally

The anchor being created by the skin is orthogonal as it is impossible to predict its value.

michael-o: OK, I see. Let's do the following:
* Support both: old and new and mark the old one in code to be removed with Doxia 2 move
* Remove old approach when plugin code is rebased to Doxia 2.0.0

The latter won't happen before end of year.

> The anchor being created by the skin is orthogonal as it is impossible to predict its value.

Correct, therefore I wouldn't even put it into consideration.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/GoalRenderer.java
```diff
@@ -0,0 +1,532 @@
...
+
+    private void renderParameterOverviewTableRow(Parameter parameter) {
+        sink.tableRow();
+        // name
+        try {
+            // link to appropriate section
+            renderTableCellWithCode(
+                    format("parameter.name", parameter.getName()),
+                    new URI(null, null, HtmlTools.encodeId(parameter.getName())));
```
michael-o, line 267: Why encode when you use the URI class, that should abstract it away no? I have the feeling that the usage of `URI` is absolutely not necessary anymore here. It maybe sufficient to pass around opaque valid strings.

kwin: Yes, the Sink impl is responsible for encoding the URL (encoding the id is necessary so that the attribute of the link and the one emitted for the anchor match). Should be fixed in https://github.com/apache/maven-plugin-tools/pull/225/commits/f6dc03408e8346922c4f350b19aef18ee4653889.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/GoalRenderer.java
```diff
@@ -0,0 +1,533 @@
...
+            }
+        }
+        // rely on the decoded link (URL encoding is done by the Sink implementation if necessary)
+        Optional<String> link = Optional.ofNullable(uri).map(u -> {
+            StringBuilder decodedLink = new StringBuilder();
+            if (u.getScheme() != null) {
+                decodedLink.append(u.getScheme()).append(":");
+            }
+            decodedLink.append(u.getSchemeSpecificPart());
+            return decodedLink.toString();
```
michael-o, line 472: Can you explain why this has to be so complex? Why not convert to string if the URI isn't null?

kwin: `URI.toString()` returns the URL-encoded String which is 
a) not necessary as the URL is encoded by [the ](https://github.com/apache/maven-doxia/blob/a1e97f9d47b20470c4a5b536406d8c5cb08277bc/doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java#L1595)
b) double encoding potentially leads to invalid links

Therefore it is crucial that the decoded URI (i.e. unencoded link) is passed to the Sink API!

michael-o: > `URI.toString()` returns the URL-encoded String which is a) not necessary as the URL is encoded by https://github.com/apache/maven-doxia/blob/a1e97f9d47b20470c4a5b536406d8c5cb08277bc/doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java#L1595 b) double encoding potentially leads to invalid links
> 
> Therefore it is crucial that the decoded URI (i.e. unencoded link) is passed to the Sink API!

I consider both statement as wrong. `URI#toString()` does not encode multibyte chars and there referenced statement just makes sure that the attribute value is valid for XML, the actual content is opaque and may be anything, even `javascript:void()`.

kwin: URI encodes parameters passed to its constructor for all cases except for the single string argument. Look at the impl. Also look at the description of the getters raw vs non-raw. In general URI internally always stores RFC compliant parts (i.e. properly encoded ones). When using toString() you would URI encode twice! Once via URI itself and the second time via the HTMLSink.

michael-o: > URI encodes parameters passed to its constructor for all cases except for the single string argument. Look at the impl. Also look at the description of the getters raw vs non-raw. In general URI internally always stores RFC compliant parts (i.e. properly encoded ones). When using toString() you would URI encode twice! Once via URI itself and the second time via the HTMLSink.

Where does the sink properly URI-encodes the input? I just don't see it.

kwin: If the link is handled opaque then it always have to be RFC 2396 in order to use it with the HTML Sink, while other sinks might not need that. That would be a serious flaw in the abstraction IMHO.

michael-o: > If the link is handled opaque then it always have to be RFC 2396 in order to use it with the HTML Sink, while other sinks might not need that. That would be a serious flaw in the abstraction IMHO.

RFC 2396 is dead for a link time so is `URI` broken. The Sink API is backend-agnostic. It cannot know what you are passing. It might be a flaw in the design, but that's the way it is. *You* have to provide valid input.

michael-o:
See also https://github.com/checkstyle/checkstyle/pull/13016

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/GoalRenderer.java
kwin: Thanks for the pointer, I hopefully clarified this a bit with https://github.com/apache/maven-doxia/pull/176/files. I am gonna rework it now to use `URI.toASCIIString` to return the URI with proper encoding.

kwin: Done in https://github.com/apache/maven-plugin-tools/pull/225/commits/1696e357ee40011660c39bfff80d58fa8dfc4ec8.

michael-o (approved):
This looks good now to me.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/GoalRenderer.java
```diff
@@ -0,0 +1,526 @@
...
+        String since = StringUtils.defaultIfEmpty(parameter.getSince(), "-");
+        renderTableCellWithCode(since);
+
+        // description
+        sink.tableCell();
+        String description;
+        String context = "Parameter " + parameter.getName() + " in goal " + descriptor.getGoal();
+        if (StringUtils.isNotEmpty(parameter.getDeprecated())) {
+            String deprecated = getXhtmlWithValidatedLinks(parameter.getDescription(), context);
+            description = format("parameter.deprecated", deprecated);
```
slawekjaranowski, line 279: @kwin @michael-o
Is it intended - we check parameter.getDeprecated() but use parameter.getDescription()?

We lost here deprecated reason ... only deprecated label with standard description is printed.

kwin: That is a copy-paste mistake.

michael-o: The removed Xdoc generator had it right. @slawekjaranowski Please file an issue. We need to fix this.

jira-importer:
Resolve #749

PR #226, opened 2023-09-08
[MPLUGIN-482] Consider implicit module name for javadoc URL calculation

kwin:
Add example JDK21 generated javadoc


maven-plugin-tools-api/pom.xml
```diff
@@ -125,4 +130,30 @@
...
+            <artifactId>maven-failsafe-plugin</artifactId>
+            <executions>
+              <execution>
+                <goals>
+                  <goal>integration-test</goal>
+                  <goal>verify</goal>
+                </goals>
+                <configuration>
+                  <!-- don't run ITs against all (non-LTS) JDK javadocs -->
+                  <excludedGroups>nonLtsJavadocs</excludedGroups>
```
slawekjaranowski, line 149: If we here exclude tests with this tag - where such tests will be executed?

kwin: only if you override that parameter or execute that IT in your IDE

kwin: always testing against all ever released JDKs just takes too much time. It is rather advisable to execute that IT manually in case something was fundamentally changed about link creation which might affect all versions.

kwin: Any other concerns @slawekjaranowski?

slawekjaranowski: ok. I see test are execute - only not the non lst are extracted to other test

slawekjaranowski (approved):

jira-importer:
Resolve #790

PR #227, opened 2023-09-25
[MNG-7416] Simplify Boolean expressions and returns

timtebeek:
@elharo since there's a long tail of projects that can still pick up diamond operators, I've started to combine those with the low volume simplify conditional expression. 
- https://issues.apache.org/jira/browse/MNG-6847
- https://issues.apache.org/jira/browse/MNG-7416

Figured that'd strike the right balance between still delivering value, while also minimizing PRs and reviews.

Use this link to re-run the recipe: https://app.moderne.io/recipes/builder/SEvWu02zw?organizationId=QXBhY2hlIE1hdmVu

### DIFF
--- a/maven-plugin-tools-java/src/main/java/org/apache/maven/tools/plugin/extractor/javadoc/JavaJavadocMojoDescriptorExtractor.java
+++ b/maven-plugin-tools-java/src/main/java/org/apache/maven/tools/plugin/extractor/javadoc/JavaJavadocMojoDescriptorExtractor.java
@@ -510,7 +510,7 @@ public class JavaJavadocMojoDescriptorExtractor extends AbstractLogEnabled
         if (superClass != null) {
             rawParams = extractFieldParameterTags(superClass);
         } else {
-            rawParams = new TreeMap<String, JavaField>();
+            rawParams = new TreeMap<>();
         }
 
         for (JavaField field : javaClass.getFields()) {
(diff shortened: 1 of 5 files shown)


maven-plugin-tools-java/src/main/java/org/apache/maven/tools/plugin/extractor/javadoc/JavaJavadocMojoDescriptorExtractor.java
```diff
@@ -510,7 +510,7 @@ private Map<String, JavaField> extractFieldParameterTags(JavaClass javaClass) {
         if (superClass != null) {
             rawParams = extractFieldParameterTags(superClass);
         } else {
-            rawParams = new TreeMap<String, JavaField>();
+            rawParams = new TreeMap<>();
```
elharo, line 513: Looking at this one, I realize the diamond operator with no params is more useful for the case where it's declared on the same line. 

TreeMap<String, JavaField> rawParams = new TreeMap<String, JavaField>();

does seem redundant

rawParams = new TreeMap<String, JavaField>();

is not. Haven't noticed that before.

elharo (approved):

maven-plugin-tools-java/src/main/java/org/apache/maven/tools/plugin/extractor/javadoc/JavaJavadocMojoDescriptorExtractor.java
elharo: Doesn't matter that much. The variable declaration wasn't in view by default in Github code review. I do think there's a real tendency for devs to argue that, "In my tool it's obvious" when advocating for more compact, less redundant syntax without really considering how it looks in other environments. I often hear folks claim that IDEs make certain constructs unnecessary. We can be really myopic about that, even to the point of saving a few milliseconds of typing right now at the cost of hours of debugging to our future selves. The real cost here is in lambdas and method chaining, which are extremely opaque and hard to read, all to save the trivial effort of typing a few more local variables.

timtebeek: I'll keep an eye out of future PRs for this pattern, although I think we've gone through the vast majority by now. There's a  couple projects left with one or two cases, which I hope to pick up along with other more meaningful changes.

PR #236, opened 2023-11-05
[MNG-5695] document Maven 3.2.5+ scoped components usage

hboutemy:
clarify something which precise scope was unclear until now (at least to me)
=> update https://maven.apache.org/plugin-tools/maven-plugin-tools-annotations/

### DIFF
--- a/maven-plugin-tools-annotations/src/site/apt/index.apt
+++ b/maven-plugin-tools-annotations/src/site/apt/index.apt
@@ -85,7 +85,7 @@ public class MyMojo
     @Parameter( name = "parameter",
                 alias = "myAlias",
                 property = "a.property",
-                defaultValue = "an expression, possibly with ${variables}",
+                defaultValue = "an expression, possibly with ${variables} and pseudo-parameter expressions ${project.xxx.yyy}",
                 readonly = <false|true>,
                 required = <false|true> )
     private String parameter;
@@ -94,27 +94,42 @@ public class MyMojo
                 hint = "..." )
     private MyComponent component;
 
+    // pseudo-parameters (marked read-only) permitting injection of Maven build context objects
     // sample objects taken from Maven API through PluginParameterExpressionEvaluator
+    //     https://maven.apache.org/ref/current/maven-core/apidocs/org/apache/maven/plugin/PluginParameterExpressionEvaluator.html
+    // plugins targetting Maven 3.2.5+ (after MNG-5695) should not use these pseudo-parameters any more,
+    // but @Component and Maven APIs to get better compiler-time checks
 
-    @Parameter( defaultValue = "${session}", readonly = true )
+    // @Parameter( defaultValue = "${session}", readonly = true )
+    @Component // since Maven 3.2.5, thanks to MNG-5695
     private MavenSession session;
 
-    @Parameter( defaultValue = "${project}", readonly = true )
+    // @Parameter( defaultValue = "${project}", readonly = true )
+    @Component // since Maven 3.2.5, thanks to MNG-5695
     private MavenProject project;
 
-    @Parameter( defaultValue = "${mojoExecution}", readonly = true )
-    private MojoExecution mojo;
+    // @Parameter( defaultValue = "${mojoExecution}", readonly = true )
+    @Component // since Maven 3.2.5, thanks to MNG-5695
+    private MojoExecution mojoExecution;
+
+    @Parameter( defaultValue = "${reactorProjects}", readonly = true )
+    // prefer using session.getProjects()
+    private List<MavenProject> reactorProjects;
 
     @Parameter( defaultValue = "${plugin}", readonly = true ) // Maven 3 only
+    // prefer using mojoExecution.getMojoDescriptor()
     private PluginDescriptor plugin;
 
     @Parameter( defaultValue = "${settings}", readonly = true )
+    // prefer using session.getSettings()
     private Settings settings;
 
     @Parameter( defaultValue = "${project.basedir}", readonly = true )
+    // prefer using project.getBasedir()
     private File basedir;
 
     @Parameter( defaultValue = "${project.build.directory}", readonly = true )
+    // prefer using project.getBuild().getDirectory()
     private File target;
 
     /**
@@ -144,3 +159,10 @@ public class MyMojo
 
  * {{{/ref/current/maven-core/apidocs/org/apache/maven/plugin/PluginParameterExpressionEvaluator.html}PluginParameterExpressionEvaluator}},
    used to evaluate plugin parameters values during Mojo configuration,
+
+ * pseudo parameters:
+
+   * <<<PluginParameterExpressionEvaluator>>> {{{https://maven.apache.org/ref/current/maven-core/apidocs/org/apache/maven/plugin/PluginParameterExpressionEvaluator.html}javadoc}} /
+   {{{https://maven.apache.org/ref/current/maven-core/xref/org/apache/maven/plugin/PluginParameterExpressionEvaluator.html}source}}
+
+   * {{{https://issues.apache.org/jira/browse/MNG-5695}MNG-5695}}: scoped objects added to Guice/Sisu in {{{https://maven.apache.org/ref/current/maven-core/}maven-core}} 3.2.5


maven-plugin-tools-annotations/src/site/apt/index.apt
```diff
@@ -94,27 +94,42 @@ public class MyMojo
...
 
+    // pseudo-parameters (marked read-only) permitting injection of Maven build context objects
     // sample objects taken from Maven API through PluginParameterExpressionEvaluator
+    //     https://maven.apache.org/ref/current/maven-core/apidocs/org/apache/maven/plugin/PluginParameterExpressionEvaluator.html
+    // plugins targetting Maven 3.2.5+ (after MNG-5695) should not use these pseudo-parameters any more,
+    // but @Component and Maven APIs to get better compiler-time checks
 
-    @Parameter( defaultValue = "${session}", readonly = true )
+    // @Parameter( defaultValue = "${session}", readonly = true )
+    @Component // since Maven 3.2.5, thanks to MNG-5695
```
slawekjaranowski, line 104: We have a warning for such components ...

https://github.com/apache/maven-plugin-tools/blob/master/maven-plugin-tools-api/src/main/java/org/apache/maven/tools/plugin/util/PluginUtils.java#L61

and Component is replaced by expression
https://github.com/apache/maven-plugin-tools/blob/master/maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/JavaAnnotationsMojoDescriptorExtractor.java#L791

gnodet: We were discussing and leaning to the exact opposite.
Keep `@Parameter` for things injected from the config and `@Component/@Inject` for injecting beans, including those from the session (Session) and mojoExecution scopes (MojoExecution, Project)...

michael-o: > We were discussing and leaning to the exact opposite. Keep `@Parameter` for things injected from the config and `@Component/@Inject` for injecting beans, including those from the session (Session) and mojoExecution scopes (MojoExecution, Project)...

This is exactly what I remember. Use `@Parameter` exactly for this case.

michael-o (changes requested):

maven-plugin-tools-annotations/src/site/apt/index.apt
gnodet: > > We were discussing and leaning to the exact opposite. Keep `@Parameter` for things injected from the config and `@Component/@Inject` for injecting beans, including those from the session (Session) and mojoExecution scopes (MojoExecution, Project)...
> 
> This is exactly what I remember. Use `@Parameter` exactly for this case.

But those components can also be injected with `@Component`, so this is weird to warn about something that is supported by sisu (because those are actually bound to scopes), while pushing users to a _custom_ solution.  Why only a few components would be accessible by name then ?
I would agree if we had a single annotation for both use cases, but that's not the case in Maven 3.  I suppose we could try for the v4 api though.

michael-o: > > > We were discussing and leaning to the exact opposite. Keep `@Parameter` for things injected from the config and `@Component/@Inject` for injecting beans, including those from the session (Session) and mojoExecution scopes (MojoExecution, Project)...
> > 
> > 
> > This is exactly what I remember. Use `@Parameter` exactly for this case.
> 
> But those components can also be injected with `@Component`, so this is weird to warn about something that is supported by sisu (because those are actually bound to scopes), while pushing users to a _custom_ solution. Why only a few components would be accessible by name then ? I would agree if we had a single annotation for both use cases, but that's not the case in Maven 3. I suppose we could try for the v4 api though.

I agree that this is inconsistent. If both are valid and are functionally identical then the warning is warong.

hboutemy: and do we now agree that using `@Component` is better than using `@Parameter( defaultValue = "${session}", readonly = true )` (because smaller and moreover we're sure there is no typo in the default value expression)?

michael-o: > and do we all now agree that using `@Component` is better than using `@Parameter( defaultValue = "${session}", readonly = true )` (because smaller and moreover we're sure there is no typo in the default value expression)?
> 
> if we agree, the warning could be to promote switch to `@Component`...

From my PoV, is this cannot be provided by POM or CLI means and it is unambiguous, then yes.

hboutemy: reading git blame, I see that in the past, we (I :) ) implemented in plugin-tools a hack to replace `@Component` with an expression in the plugin descriptor: https://issues.apache.org/jira/browse/MPLUGIN-204

I don't remember when this was done, if it was before Maven 3.2.5 MNG-5695, but with Maven 3.2.5, this hack is not useful for `MavenSession`, `MavenProject` nor `MojoExecution`: it remains useful for a few other types...

hboutemy:
now that we've found back the MPLUGIN-204 hack, in the proposed update:
```
@Component // since Maven 3.2.5, thanks to MNG-5695
```
we could perhaps drop the comment on Maven 3.2.5 because the plugin transforms `@Component` with `@Parameter`, then it works with any Maven release
the only improvement that could be useful is to NOT transform but keep a real Plexus component if the plugin is targetting 3.2.5+: I don't know if this is feasible

hboutemy:
in MPLUGIN-204 we advised using `@Component` (that would use the trick to transform to parameters in the plugin descriptor XML), but later we advised back: https://github.com/apache/maven-plugin-tools/commit/f8a946df81176cfdf4be2f37c04b881b961d6994
see https://issues.apache.org/jira/browse/MPLUGIN-257

gnodet:
> in [MPLUGIN-204](https://issues.apache.org/jira/browse/MPLUGIN-204) we advised using `@Component` (that would use the trick to transform to parameters in the plugin descriptor XML), but later we advised back: [f8a946d](https://github.com/apache/maven-plugin-tools/commit/f8a946df81176cfdf4be2f37c04b881b961d6994) see https://issues.apache.org/jira/browse/MPLUGIN-257

That's exactly what I don't understand.  Since they are bound to a scope in plexus, they _are_ plexus components, even if they have not been instantiated by plexus.

maven-plugin-tools-annotations/src/site/apt/index.apt
cstamas: IMHO there is no problem when JSR330 being used. Problem happens when Mojo APIs (java5 anno or javadoc annos are being used).

hboutemy:
before Maven 3.2.5, they were not Plexus components: then implementation in plexus-tools was a hack to replace with parameter in plugin descriptor XML

hboutemy:
on latest discussions on Slack (I hate non-shared Slack diuscussions), there is also the `@Inject` options that IIUC completely avoids the plugin descriptor intermediate= `<requirement` in https://maven.apache.org/ref/3.9.5/maven-plugin-api/plugin.html

I don't know if it works, never tried

michael-o (approved):

cstamas (approved):

michael-o:
Are we good to merge this? We have two approvals.

slawekjaranowski (changes requested):
We also should remove replacements and worning form code:

https://github.com/apache/maven-plugin-tools/blob/master/maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/JavaAnnotationsMojoDescriptorExtractor.java#L791

changing only documentation users will be confused when use it

michael-o:
> We also should remove replacements and worning form code:
> 
> https://github.com/apache/maven-plugin-tools/blob/master/maven-plugin-tools-annotations/src/main/java/org/apache/maven/tools/plugin/extractor/annotations/JavaAnnotationsMojoDescriptorExtractor.java#L791
> 
> changing only documentation users will be confused when use it

Right, it needs to be consistent. @cstamas, your turn.

michael-o:
I took the stab now and change the code: #253. Both PRs should go together in the next release.

PR #239, opened 2023-11-10
[MPLUGIN-450] Require goalPrefix to be valid

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MPLUGIN-450


jira-importer:
Resolve #767

PR #250, opened 2023-12-25
[MPLUGIN-499] Add deprecate description in parameters table

slawekjaranowski:


michael-o (changes requested):
The full stop after "Deprecated" is on purpose because `javadoc(1)` does the same: See: https://codehaus-plexus.github.io/plexus-utils/apidocs/org/codehaus/plexus/util/cli/Commandline.html#createArgument(). We wanted to resemblance.

slawekjaranowski:
ok, next version

michael-o:
Checking...

michael-o:
Looks better now. I see one difference:
Before:
grafik

After:
grafik

The full stop is gone. Is that on purpose?

slawekjaranowski:
Missing dot on list of goals is by accident. The same resources is used on goal details page:

image

It was:

```
Deprecated.:
```

slawekjaranowski:
But it is a next issue on list of goals - plugin-info.htm when goal is deprecated we don't have a description.

michael-o:
> Missing dot on list of goals is by accident. The same resources is used on goal details page:
> 
> image
> 
> It was:
> 
> ```
> Deprecated.:
> ```

Yes, one fixed the other one broken :-D Can you address?

michael-o:
> But it is a next issue on list of goals - plugin-info.htm when goal is deprecated we don't have a description.

This isn't right. Deprecation should be first, line break, description.

slawekjaranowski:
Now it will be as:
image

michael-o:
> Now it will be as: image

Good. There should be a space between "Deprecated." and the actual text.

Let me test meanwhile.

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginOverviewRenderer.java
```diff
@@ -144,15 +145,26 @@ private void renderGoalsSection(boolean hasMavenReport) {
...
+            sink.tableCell();
             if (StringUtils.isNotEmpty(mojo.getDeprecated())) {
-                description = "<strong>" + getI18nString("goal.deprecated") + "</strong> " + mojo.getDeprecated();
-            } else if (StringUtils.isNotEmpty(mojo.getDescription())) {
+                sink.division();
+                sink.inline(SinkEventAttributeSet.Semantics.STRONG);
+                sink.text(getI18nString("goal.deprecated"));
+                sink.text(".");
+                sink.inline_();
+                sink.rawText(mojo.getDeprecated());
```
michael-o, line 155: Should be preceded with a space.

michael-o (changes requested):

maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginOverviewRenderer.java
slawekjaranowski: done

michael-o (approved):

jira-importer:
Resolve #805

PR #271, opened 2024-03-20
switch from png+imagemap to svg

hboutemy:
same test as https://github.com/apache/maven-site/pull/504

result published to https://maven.apache.org/plugin-tools-archives/plugin-tools-LATEST/
vs initial png https://maven.apache.org/plugin-tools-archives/plugin-tools-3.11.0/


jira-importer:
Resolve #816

PR #273, opened 2024-03-23
use @Component instead of @Parameter when possible

hboutemy:


slawekjaranowski:
For it we should m-plugin-tolls in version 3.11.0 😄

PR #274, opened 2024-03-23
[MPLUGIN-510] group history per common requirements

hboutemy:
instead of listing a long list of plugin releases, each on their own row, even if they share same system requirements, they are now grouped in a single row marked "from x to y"


jira-importer:
Resolve #824

PR #280, opened 2024-04-29
[MPLUGIN-519][MPLUGIN-520] Parent POM 42, prerequisite of 3.6.3, get rid of maven-compat

cstamas:
Update parent POM to 42, up prerequisite to 3.6.3. As parent 42 removes implicit (sisu) indexing, added explicitly indexing to all modules, as I think they all need it (may be wrong here).

The Mojo `AddPluginArtifactMetadataMojo` pulls in classes that have transitive (classes) coming from maven-compat. In fact, this mojo is not required at all since Maven 3.9.0 as resolver handles all transparently.

---

https://issues.apache.org/jira/browse/MPLUGIN-519
https://issues.apache.org/jira/browse/MPLUGIN-520

### DIFF
--- a/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/metadata/AddPluginArtifactMetadataMojo.java
+++ b/maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/metadata/AddPluginArtifactMetadataMojo.java
@@ -18,10 +18,8 @@
  */
 package org.apache.maven.plugin.plugin.metadata;
 
-import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.repository.metadata.ArtifactRepositoryMetadata;
 import org.apache.maven.artifact.repository.metadata.GroupRepositoryMetadata;
-import org.apache.maven.artifact.repository.metadata.Versioning;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.descriptor.PluginDescriptor;
@@ -30,6 +28,10 @@ import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.project.MavenProject;
+import org.apache.maven.rtinfo.RuntimeInformation;
+import org.eclipse.aether.util.version.GenericVersionScheme;
+import org.eclipse.aether.version.InvalidVersionSpecificationException;
+import org.eclipse.aether.version.VersionScheme;
 
 /**
  * Inject any plugin-specific
@@ -69,6 +71,11 @@ public class AddPluginArtifactMetadataMojo extends AbstractMojo {
     @Parameter(defaultValue = "false", property = "maven.plugin.skip")
     private boolean skip;
 
+    @Component
+    private RuntimeInformation runtimeInformation;
+
+    private final VersionScheme versionScheme = new GenericVersionScheme();
+
     /** {@inheritDoc} */
     @Override
     public void execute() throws MojoExecutionException {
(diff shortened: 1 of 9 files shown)


maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/metadata/AddPluginArtifactMetadataMojo.java
```diff
@@ -18,10 +18,8 @@
  */
 package org.apache.maven.plugin.plugin.metadata;
 
-import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.repository.metadata.ArtifactRepositoryMetadata;
 import org.apache.maven.artifact.repository.metadata.GroupRepositoryMetadata;
```
gnodet, line 22: Those two imports should be removed, they are referenced from the javadoc and they should be moved to LegacySupport.

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/metadata/AddPluginArtifactMetadataMojo.java
```diff
@@ -42,8 +42,8 @@
  * <li>to define plugin mapping in the group</li>
  * </ol>
  *
- * @see ArtifactRepositoryMetadata
- * @see GroupRepositoryMetadata
+ * @see org.apache.maven.artifact.repository.metadata.ArtifactRepositoryMetadata
+ * @see org.apache.maven.artifact.repository.metadata.GroupRepositoryMetadata
```
michael-o, line 46: These links will not work if Javadoc does not find the deps on the classpath.

gnodet: Maven-compat is still in the dependencies afaik.  It has to be until we actually completely remove this mojo.

michael-o (approved):
This works for me.

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/metadata/AddPluginArtifactMetadataMojo.java
```diff
@@ -69,25 +69,33 @@ public class AddPluginArtifactMetadataMojo extends AbstractMojo {
...
+                getLog().info("This Mojo is not used in Maven version 3.9.0 and above");
+                return;
+            }
+        } catch (InvalidVersionSpecificationException e) {
+            // not happening with generic
+            throw new MojoExecutionException(e);
+        }
 
-        projectArtifact.addMetadata(groupMetadata);
+        LegacySupport.execute(project, getGoalPrefix());
```
slawekjaranowski, line 98: We should be careful with it .... probably it is used by Nexus Staging Plugin
https://issues.apache.org/jira/browse/MPLUGIN-384

cstamas: I consider nexus-staging-maven-plugin legacy, simply put, people should avoid using it... all the benefits it (may) add, is that you don't have to log into Nx UI and press "close" button. Simply put, not worth if you contrast that to legacy burden it brings and have.

michael-o: I fully agree with @cstamas and ignore it.

michael-o:
This will constitute a great 3.13.0 release.

michael-o (approved):

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/metadata/AddPluginArtifactMetadataMojo.java
slawekjaranowski: 👍

slawekjaranowski: We will see how many people use nexus-staging

slachiewicz (approved):

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/metadata/AddPluginArtifactMetadataMojo.java
slawekjaranowski: Is it materialized .... https://issues.apache.org/jira/browse/MNG-8121 ❓

cstamas: This change is unrelated. What happened is that those users had no prefix set (ever), so remote metadata has no prefix either, and now that they upgraded, and forced to set plugin prefix (in latest m-p-p), NPE happens.

PR #285, opened 2024-05-23
[MPLUGIN-525] Remove ant and beanshell support

gnodet:
- **[MPLUGIN-523] Switch to version 4.0.0-SNAPSHOT**
- **[MPLUGIN-525] Remove ant and beanshell support**


michael-o:
Is this Maven 4 only?

gnodet:
> Is this Maven 4 only?

Yes, I've created a [3.x branch](https://github.com/apache/maven-plugin-tools/tree/3.x) to support Maven 3 plugins and master will support Maven 4 plugins. The goal is to split the plugins to ease maintenance, as the code will ultimately be quite different.  See https://issues.apache.org/jira/browse/MPLUGIN-523. We should also drop the Maven 4.x support from the 3.x branch.

michael-o:
> > Is this Maven 4 only?
> 
> Yes, I've created a [3.x branch](https://github.com/apache/maven-plugin-tools/tree/3.x) to support Maven 3 plugins and master will support Maven 4 plugins. The goal is to split the plugins to ease maintenance, as the code will ultimately be quite different. See https://issues.apache.org/jira/browse/MPLUGIN-523. We should also drop the Maven 4.x support from the 3.x branch.

I see. There is no way to have this in one branch with common modules and module per Maven version?

PR #286, opened 2024-05-23
[MPLUGIN-508] Upgrade to Maven 4.0.0-beta-3

gnodet:
- **[MPLUGIN-508] Upgrade to 4.0.0-beta-3**


michael-o:
Please hold this one off until #287 and 3.13.1 has been cut. I want to clean up first before we cut 3.x.

maven-plugin-plugin/src/it/v4api/src/main/java/org/apache/maven/its/v4api/FirstMojo.java
```diff
@@ -38,15 +37,12 @@
...
  * @since 1.2
  */
-@Mojo(
-        name = "first",
-        requiresDependencyResolution = ResolutionScope.TEST,
-        defaultPhase = LifecyclePhase.INTEGRATION_TEST)
-@Execute(phase = LifecyclePhase.GENERATE_SOURCES, lifecycle = "cobertura")
+@Mojo(name = "first", defaultPhase = "integration-test")
+@Execute(phase = "generate-sources", lifecycle = "cobertura")
 public class FirstMojo implements org.apache.maven.api.plugin.Mojo {
```
michael-o, line 46: Is there a reason why other args to `@Mojo` are gone?

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/DescriptorGeneratorMojo.java
```diff
@@ -78,6 +90,12 @@ public class DescriptorGeneratorMojo extends AbstractGeneratorMojo {
     @Parameter(defaultValue = "${project.build.outputDirectory}/META-INF/maven", readonly = true)
     private File outputDirectory;
 
+    /**
+     * The directory where the generated class files will be put.
+     */
+    @Parameter(defaultValue = "${project.build.outputDirectory}", readonly = true)
+    private File classOutputDirectory;
```
michael-o, line 97: `classesOutputDirectory`?

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/DescriptorGeneratorMojo.java
```diff
@@ -367,6 +390,141 @@ public void generate() throws MojoExecutionException {
...
+                    //                        diBeans.add(className);
+                    //                    }
+                }
+            }
+            Path path = outputDirectory.toPath().resolve("org.apache.maven.api.di.Inject");
+            if (diBeans.isEmpty()) {
+                Files.deleteIfExists(path);
+            } else {
+                try (CachingWriter w = new CachingWriter(path, StandardCharsets.UTF_8)) {
+                    String content = diBeans.stream().collect(Collectors.joining("\n", "", "\n"));
```
michael-o, line 449: `System.lineSeparator()`?

michael-o (changes requested):

maven-plugin-plugin/src/it/v4api/src/main/java/org/apache/maven/its/v4api/FirstMojo.java
gnodet: Maven 4 mojos have to be thread safe.
In maven 3, the `requiresDependencyResolution` and `requiresDependencyCollection` are currently used to collect or resolve the project dependencies in a given scope.  Those artifacts are then made available by Maven by `MavenProject.getArtifacts()`.  This brings several problems: for a single project, two mojos can not be run concurrently, and as we've seen, some mojos even require runtime + test scope resolution, and that cannot be provided.   In 4.x, mojos will have to use the API to retrieve collect/resolve the dependencies.  Hopefully the result can be cached in the session.  So the latest 4.x release did remove those two bits.  However, not having this information sounds wrong, as this would be needed in at least two scenarii:
 * it would be nice to be able to eagerly download dependencies at the beginning of a given build, and we'd need to know which scopes are needed for a given project, this could be done by analysing the build plan
 * if we experiment a bit more with the [fully concurrent builder](https://github.com/apache/maven/pull/1429), we need this information in order to know which goals depend on which projects in the reactor, so that the goals can be scheduled correctly
So I think we'll need to add those back somehow, but they would be informational.  Another possibility would be to allow injecting the resolution/collection result in the mojo.

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/DescriptorGeneratorMojo.java
gnodet: Fixed

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/DescriptorGeneratorMojo.java
gnodet: Fixed

maven-plugin-plugin/src/it/v4api/src/main/java/org/apache/maven/its/v4api/FirstMojo.java
gnodet: I raised https://issues.apache.org/jira/browse/MNG-8134

michael-o: > Maven 4 mojos have to be thread safe, so that annotation is gone.

You cannot enforce that right? Only by contract...

maven-plugin-plugin/src/it/skip/src/main/java/org/apache/maven/plugin/coreit/FirstMojo.java
```diff
@@ -18,47 +18,15 @@
...
- *
- * @goal first
- * @requiresDependencyResolution test
- * @phase integration-test
- * @execute phase="generate-sources" lifecycle="cobertura"
- * @deprecated Don't use!
- * @since 1.2
  */
+@Mojo(name = "test")
 public class FirstMojo extends AbstractMojo {
```
michael-o, line 29: This is now completely different, how?

maven-plugin-plugin/src/it/v4api/invoker.properties
```diff
@@ -15,4 +15,5 @@
 # specific language governing permissions and limitations
 # under the License.
 
+invoker.java.version = 17+
```
michael-o, line 18: Confusing if the entire project require Java 17, why here especially?

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/DescriptorGeneratorMojo.java
```diff
@@ -48,8 +54,14 @@
 import org.apache.maven.tools.plugin.scanner.MojoScanner;
 import org.codehaus.plexus.component.repository.ComponentDependency;
 import org.codehaus.plexus.util.ReaderFactory;
+import org.codehaus.plexus.util.io.CachingOutputStream;
+import org.codehaus.plexus.util.io.CachingWriter;
+import org.objectweb.asm.*;
```
michael-o, line 59: Is this necessary?

michael-o (changes requested):

maven-plugin-plugin/src/it/v4api/src/main/java/org/apache/maven/its/v4api/FirstMojo.java
gnodet: > > Maven 4 mojos have to be thread safe, so that annotation is gone.
> 
> You cannot enforce that, right? Only by contract...

Right.

maven-plugin-plugin/src/it/v4api/invoker.properties
gnodet: The entire project is JDK 8.  The mojos run on Maven 3.x + JDK 8 if you want to build Maven 3.x plugins.  If you build a Maven 4.x plugin, you'll need JDK 17.  That's the case for this IT.  Other ITs are unchanged.

The migration to Maven 4 only should imho happen at the same time as m-plugin-p is migrated to be a Maven 4 plugin.

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/DescriptorGeneratorMojo.java
gnodet: The imports are used to generate the factories for Maven 4.x mojos.  They are not using plexus, as they are booted by the new DI framework in Maven.  But this framework recognises beans annotated with `@org.apache.maven.api.di.Named`, which is not the case for mojos.  To avoid forcing the user to add this annotation, a factory (annotated with the needed annotation) is generated by the plugin.  And to avoid having to do another compile phase, the byte code is directly generated instead of generating source code.  Those new imports are needed for this feature.

maven-plugin-report-plugin/src/it/mplugin-319_report-since/src/main/java/org/MyMojo.java
```diff
@@ -19,41 +19,32 @@
...
- * @requiresOnline
- * @inheritByDefault false
- * @execute phase="compile"
- * @aggregator
  * @since 1.0
  *
  * @deprecated You don't use test goals, do you?
  */
+@Mojo(name = "noop")
 public class MyMojo extends AbstractMojo {
```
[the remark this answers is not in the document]
gnodet: Why ? if not specified, the plugin won't have a default phase, but that's not a problem here.  We don't even use this plugin, we're only building it.

maven-plugin-plugin/src/main/java/org/apache/maven/plugin/plugin/DescriptorGeneratorMojo.java
michael-o: Accepted.

maven-plugin-report-plugin/src/it/mplugin-319_report-since/src/main/java/org/MyMojo.java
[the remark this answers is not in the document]
michael-o: The previous config had a lot of stuff:
```
 * @goal noop
 * @phase process-sources
 * @requiresDependencyResolution test
 * @requiresDirectInvocation true
 * @requiresOnline
 * @inheritByDefault false
 * @execute phase="compile"
 * @aggregator
 ```

Plain `@Mojo` isn't an equivalent, is it?

File: maven-plugin-report-plugin/src/main/java/org/apache/maven/plugins/plugin/descriptor/EnhancedPluginDescriptorBuilder.java [hunk 8]
```diff
@@ -43,7 +42,7 @@
  * Reads enhanced plugin.xml files as generated by
  * {@link PluginDescriptorFilesGenerator} and used by {@link PluginReport}.
  * Populates the slightly extended {@link Parameter} object {@link EnhancedParameterWrapper}.
- * In addition populates all (optional) elements added after Maven Plugin API 3.2.5.
+ * In addition, populates all (optional) elements added after Maven Plugin API 3.2.5.
```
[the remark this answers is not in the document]
michael-o: Alright, let's keep the wording as-is.

maven-plugin-report-plugin/src/it/mplugin-319_report-since/src/main/java/org/MyMojo.java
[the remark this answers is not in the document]
gnodet: No, but again, those are not tested.  I think the one creating the original test just copy/pasted an existing one.

maven-plugin-plugin/src/it/skip/src/main/java/org/apache/maven/plugin/coreit/FirstMojo.java
gnodet: Those mojos are just built and used to verify a few specific things depending on each test.  The annotations are usually not used or tested.

michael-o (approved):
I have no further objections, though two others should have a look as well.

PR #296, opened 2024-06-13
Run the v4 mojo

gnodet:


gnodet:
This PR currently fail as expected, as the Maven 4 mojo cannot be run on Maven 3.
Not sure how to fix that...

PR #317, opened 2024-08-10
[MPLUGIN-530] Deprecate o.a.m.plugins.annotations.Component

kwin:
Recommend using JSR 330 annotations instead


slachiewicz (approved):

kwin:
Probably after MPLUGIN-523 this should only target branch `maven-plugin-tools-3.x`

PR #319, opened 2024-08-15
Remove Maven3 remnants

kwin:
(WIP)


kwin:
@gnodet Is this going in the right direction? A lot of work still to be done....

gnodet:
> @gnodet Is this going in the right direction? A lot of work still to be done....

I think so.  To port plugins to the v4 api, basically they should not have any dependency coming from maven other than org.apache.maven.api packages.

ascopes:
Will this include the fixing of warnings in code produced by maven-plugin-plugin for v4?

As an example, generated HelpMojo classes now fail in projects with -Werror as they still access deprecated APIs such as .getLog().

PR #345, opened 2024-12-13
Update document to use Guice constructor injection

elharo:


maven-plugin-tools-annotations/src/site/apt/index.apt
```diff
@@ -132,6 +110,27 @@ public class MyMojo
...
+
+    private final MyComponent component;
+
+    private final MavenSession session;
+
+    private final MavenProject project;
+
+    private final MojoExecution mojoExecution;
+
+    public MyMojo(MyComponent component, MavenSession session, MavenProject project, MojoExecution mojoExecution) {
```
slachiewicz, line 127: @Inject annotation missing

elharo: fixed

slawekjaranowski (approved):
Should be also updated for 3.x branch?

elharo:
> Should be also updated for 3.x branch?

Do we publish both?

PR #353, opened 2025-02-12
[MPLUGIN-541] Rename report from "Plugin Documentation" to "Plugin

kwin:
Details"


michael-o (changes requested):
You need to fix the other bundles as well otherwise they will have wrong labels.

kwin:
> You need to fix the other bundles as well otherwise they will have wrong labels.

Those are outdated anyways (just compare the amount of keys in the global/en one with the other languages) and I don't speak those languages so don't dare to just put them in. Instead of deleting I would suggest just leaving them stale.

michael-o (approved):

jira-importer:
Resolve #850

PR #356, opened 2025-02-17
Simply GitHub verify action

slawekjaranowski:


.github/workflows/maven-verify.yml
```diff
@@ -26,7 +26,5 @@ jobs:
     name: Verify
     uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v4
     with:
-      maven-matrix: '[ "3.6.3", "3.9.9", "4.0.0-beta-4" ]'
-      jdk-matrix: '[ "8", "17", "21" ]'
```
Bukama, line 30: Planed to remove the JDK matrix too with 17 and 21 (8 is obvious)

slawekjaranowski: simply use default matrix defined in shared

Bukama (approved):

PR #401, opened 2025-05-14
Add support for Maven 4 PluginDescriptor.getRequiredJavaVersion() method

gnodet:
## Description

This PR enhances the `PluginDescriptorHelper` class to support the `getRequiredJavaVersion()` and `setRequiredJavaVersion()` methods directly on the `PluginDescriptor` class in Maven 4.

## Implementation Details

The implementation:
1. Uses reflection to check if the methods exist in the current Maven version
2. Tries to use the direct methods first if available (Maven 4)
3. Falls back to the existing wrapper approach if needed (Maven 3)

This approach ensures backward compatibility while taking advantage of the new methods in Maven 4.

## Changes
- Added static fields to hold method references using reflection
- Enhanced the existing `getRequiredJavaVersion()` method to try the direct method first
- Added a new `setRequiredJavaVersion()` method following the same pattern
- Added comprehensive JavaDoc documentation for both methods

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) with guidance from the PR author

### DIFF
--- a/maven-plugin-tools-api/src/main/java/org/apache/maven/tools/plugin/PluginDescriptorHelper.java
+++ b/maven-plugin-tools-api/src/main/java/org/apache/maven/tools/plugin/PluginDescriptorHelper.java
@@ -1,0 +1,100 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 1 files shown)


maven-plugin-tools-api/src/main/java/org/apache/maven/tools/plugin/PluginDescriptorHelper.java
slawekjaranowski, line 1: How it will be used?

Why not direct use class from Maven core ... such master branch will be dedicated to Maven 4?

gnodet: We could, but I still think being able to build a maximum of projects targeting Maven 3 with Maven 4 is a good idea.  We do want maximum compatibility when possible.
Also, if we target Maven 4, we should move to the new API, but that's more work imho.
So I'd rather to a micro release, and then start migrating this plugin to target Maven 4.

slawekjaranowski: so should we do it in branch: maven-plugin-tools-3.x ?

gnodet: > so should we do it in branch: maven-plugin-tools-3.x ?

Yes, I suppose.  Not sure we need to do it now, unless someone wants to start the migration asap though.

slawekjaranowski (approved):

PR #1005, opened 2025-10-14
Add support for Maven 4 PluginDescriptor.getRequiredJavaVersion() method

slawekjaranowski:
* Add support for Maven 4 PluginDescriptor.getRequiredJavaVersion() method

This commit enhances the PluginDescriptorHelper class to support the getRequiredJavaVersion() and setRequiredJavaVersion() methods directly on the PluginDescriptor class in Maven 4.

The implementation:
1. Uses reflection to check if the methods exist in the current Maven version
2. Tries to use the direct methods first if available (Maven 4)
3. Falls back to the existing wrapper approach if needed (Maven 3)

This approach ensures backward compatibility while taking advantage of the new methods in Maven 4.

* Add missing files

(cherry picked from commit b9393905ec66824d89b5fdfce22246dee9fe0a77)

- #401


slawekjaranowski:
@gnodet I'm not sure if we should label it as `bug` ... rather it is enhancement or maintenance

PR #1029, opened 2025-11-16
Check mkdirs() return value when creating output directory in DescriptorGeneratorMojo

MateusJunior7:
This PR addresses a RV_RETURN_VALUE_IGNORED_BAD_PRACTICE warning reported by SpotBugs in
org.apache.maven.plugin.plugin.DescriptorGeneratorMojo.generate().

The call to java.io.File.mkdirs() previously ignored its boolean return value.
The code now:

Checks whether the output directory already exists.

Calls mkdirs() only when necessary.

Throws a MojoExecutionException when directory creation fails, instead of silently ignoring the error.

After this change, the RV warning no longer appears in the SpotBugs report for this class.


Bukama:
part of #1030

PR #1146, opened 2026-08-04
Build against Maven 4.0.0-rc-6

slachiewicz:
4.0.0-rc-6 is released. This moves the Maven 4 version property to it and pins the same version explicitly in the Verify workflow, so CI no longer depends on the `maven4-version` default in maven-gh-actions-shared.

Verified: green on the fork before opening (`Verify` workflow, full matrix).

### DIFF
--- a/.github/workflows/maven-verify.yml
+++ b/.github/workflows/maven-verify.yml
@@ -27,4 +27,5 @@ jobs:
     uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v5
     with:
       maven4-enabled: true  # build by Maven 3.x and 4.x
+      maven4-version: '4.0.0-rc-6' # the same as used in project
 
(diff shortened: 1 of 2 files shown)


.github/workflows/maven-verify.yml
```diff
@@ -27,4 +27,5 @@ jobs:
     uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v5
     with:
       maven4-enabled: true  # build by Maven 3.x and 4.x
+      maven4-version: '4.0.0-rc-6' # the same as used in project
```
slawekjaranowski, line 30: I would like to consider to upgrade in shared actions

slawekjaranowski (approved):
What about jenkins?

.github/workflows/maven-verify.yml
slachiewicz: last time I've updated it globally in gha shared and then got lots of compalins, so this time repo by repo

Issue #944, opened 2025-07-06
Unable to parse generated classes during descriptor generation


ascopes:
### Affected version

3.15.1

### Bug description

I am using `org.immutables:immutables` to generate code for my Maven Plugin project.

One of the components I am using is `org.immutables:data` which generates reflective metadata for a generated class.

It seems that the maven-plugin-plugin is attempting to parse this and falling over, despite the fact it is valid code.

The line it chokes on appears to be the following:

```java
    public final Datatype.Feature<MavenProtocPlugin, @Nullable List<String>> jvmArgs_;
```
specifically at the end of the `List` identifier before the `<` that follows it.

As a result of this, I am unable to use generated code with the maven-plugin-plugin, since there is no way to ask it to skip scanning generated code -- not entirely sure why it is attempting to scan this generated code as it is not a Mojo that should be converted to documentation.

Maven: 3.9.10

Java: 17

Exception:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-plugin-plugin:3.15.1:descriptor (default-descriptor) on project protobuf-maven-plugin: Execution default-descriptor of goal org.apache.maven.plugins:maven-plugin-plugin:3.15.1:descriptor failed: syntax error @[64,68] in file:/home/ashley/code/protobuf-maven-plugin/protobuf-maven-plugin/target/generated-sources/annotations/io/github/ascopes/protobufmavenplugin/plugins/Datatypes_MavenProtocPlugin.java -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-plugin-plugin:3.15.1:descriptor (default-descriptor) on project protobuf-maven-plugin: Execution default-descriptor of goal org.apache.maven.plugins:maven-plugin-plugin:3.15.1:descriptor failed: syntax error @[64,68] in file:/home/ashley/code/protobuf-maven-plugin/protobuf-maven-plugin/target/generated-sources/annotations/io/github/ascopes/protobufmavenplugin/plugins/Datatypes_MavenProtocPlugin.java
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:333)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:316)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:212)
[...]

PR #1150, opened 2026-08-07
[MPLUGIN-421] Replace QDox with JavaParser for Javadoc extraction

### LINKED ISSUE #718: [MPLUGIN-421] descriptor goal fails when there is .java file present with content that's just a comment
**[Leszek Jadacki](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=JIRAUSER296637)** opened **[MPLUGIN-421](https://issues.apache.org/jira/browse/MPLUGIN-421?redirect=false)** and commented

Hello,

As I was learning to create own custom plugin, during development process I decided not to compile one of my Mojo's. For ease I simply commented all its content. 

When trying to build my "custom" plugin _Install_ fails when it executes _descriptor_ goal.

It seems like when there is a file in the source folder with .java extension but contains only commented lines the compiler omits the file, but the descriptor is taking it into account and fails during parsing the file with error "syntax error @[X,Y]" pointing to it's last line and column.

---

**Affects:** 3.6.4

wilx:
## Summary

Replace the annotation extractor's QDox-backed Java source model with JavaParser and its symbol solver. This fixes descriptor generation for comment-only Java source files while retaining Javadoc extraction, reference resolution, and link generation.

Fixes #718.

## Implementation

- Add a JavaParser source model that resolves project sources, reactor output directories, dependency JARs, and the complete Java class library.
- Port Mojo and parameter Javadoc extraction, member lookup, static-field values, module lookup, and internal/external link resolution from QDox to JavaParser.
- Keep directory and JAR classpath entries visible through one class loader with platform-module delegation.
- Add regression coverage for comment-only sources, mixed reactor/JAR inheritance, JDK platform modules, generated sources, and observable source-root exclusion.
- Remove the annotation extractor's direct QDox dependency and update the dependency diagram.

The user-visible result is that valid projects containing a `.java` file with only comments no longer fail during the `descriptor` goal. Existing Javadoc metadata and links continue to be generated through the new model.

## Validation

- `JAVA_HOME=/opt/jdks/latest-17 mvn spotless:apply`
- `JAVA_HOME=/opt/jdks/latest-17 mvn -Prun-its verify`
  - Maven Plugin Plugin ITs: 31 passed, 1 Maven-version-specific test skipped
  - Maven Plugin Report Plugin ITs: 12 passed

Following this checklist to help us incorporate your
contribution quickly and easily:

- [x] Your pull request should address just one issue, without pulling in other changes.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ] Each commit in the pull request should have a meaningful subject line and body.
  Note that commits might be squashed by a maintainer on merge.
- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
  This may not always be possible but is a best-practice.
- [x] Run `mvn verify` to make sure basic checks pass.
  A more thorough check will be performed on your pull request automatically.
- [x] You have run the integration tests successfully (`mvn -Prun-its verify`).

If your pull request is about ~20 lines of code you don't need to sign an
[Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
please ask on the developers list.

To make clear that you license your contribution under
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

- [x] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
- [ ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).


slachiewicz (commented):
Thanks for tackling this — the JavaParser model is a big improvement over QDox for the comment-only case, and the IT coverage is good. I pushed one commit (`bf42502`) to fix the Windows CI failure, and below are the findings from a closer review.

## CI

The run looked worse than it was: of the 15 red jobs, only **one** was a real failure — `windows-latest jdk-21-zulu` — and fail-fast cancelled the other 14 at ~2 minutes.

`JavaSourceModel.addSourceDirectory()` canonicalizes source roots with `Path.toRealPath()`, so the `IOException` message carries the canonical path, but `JavaSourceModelTest.reportsSourcePathForInvalidJava` compared it against the raw JUnit `@TempDir` path. On the Windows runners `java.io.tmpdir` is handed out in 8.3 short form (`C:\Users\RUNNER~1\...`), so the two spellings never matched. Fixed in `bf42502` by comparing against `source.toRealPath()`. Reproducible off-Windows with `-DargLine=-Djava.io.tmpdir=<a symlink>`.

With that green, a **second** failure appeared behind it — see the last section, which is not your bug.

## Confirmed findings

**1. The `URLClassLoader` is leaked whenever `parse()` fails.** `JavaSourceModel.parse()` creates `classPathLoader` (`JavaSourceModel.java:120`) *before* the source-parsing loop that throws on the first parse problem (line 134). `parse()` is called at the end of `scanJavadoc()` (`JavaAnnotationsMojoDescriptorExtractor.java:299`), i.e. before the try-with-resources at line 224 takes ownership of the returned model — so on any parse failure `close()` never runs and every dependency/reactor jar stays open for the life of the JVM. On Windows those files stay locked. Simplest fix: return the unparsed model from `scanJavadoc()` and call `sourceModel.parse()` inside the try block; that also gives a cleaner error path.

**2. Nested mojo classes silently lose all their Javadoc.** `mojoAnnotatedClasses` is keyed by ASM's `Type.getObjectType(name).getClassName()` (`MojoClassVisitor.java:118`), which keeps the binary form `pkg.Outer$Inner`. `discoverClasses()` (`JavaAnnotationsMojoDescriptorExtractor.java:485-491`) keys on JavaParser's `getFullyQualifiedName()`, which is canonical: `pkg.Outer.Inner`. So the lookup at line 316 returns `null` and `continue`s, dropping the description, `@since` and `@deprecated` with no warning. QDox's `getFullyQualifiedName()` returned the binary form, so this worked before — it is a silent regression. Normalising `$` to `.` at the lookup (or indexing types under both names) fixes it; worth an IT with an `@Mojo` declared as a nested class.

**3. `hasPackage()` consults `Package.getPackage()`** (`JavaSourceModel.java:193`). That is deprecated since Java 9 and, worse, caller-classloader-sensitive: it only sees packages already loaded by the plugin-tools realm, so whether a package-only reference resolves depends on what maven-plugin-tools itself happens to have loaded. That is non-deterministic and leaks the tool's own classpath into user descriptors. QDox consulted only its own model. Suggest restricting this to the indexed packages plus a real JDK/boot-layer package list.

**4. `javaparser-core` is used but not declared.** The module imports `com.github.javaparser.ast.*` but the pom only declares `javaparser-symbol-solver-core` (`maven-plugin-tools-annotations/pom.xml:100`). `dependency:analyze` will flag this as used-undeclared.

**5. Member-only reference lookup no longer searches nested classes** of the declaring class (`JavaClassConverterContext.java:177-182`; the old code had an explicit `getNestedClasses()` pass). `{@link #x}` where `x` lives in a nested class now degrades to a warning plus an unresolved-link comment. Low impact — the old behaviour was itself not javadoc-conformant — but it is a silent output change worth noting.

## Worth a decision

**Aborting the whole descriptor generation on the first unparseable file** (`JavaSourceModel.java:128-135`) is a meaningful behaviour change, and I would push back on it. Two things make it riskier than it looks:

- QDox skipped method bodies, so only declaration-level syntax could fail. JavaParser parses full bodies at 3.28.2's `BLEEDING_EDGE` ceiling, so any body-level syntax newer than its grammar is now fatal.
- `excludedScanDirectories` is only applied to the project's own compile source roots (`JavaAnnotationsMojoDescriptorExtractor.java:552-561`). The sources-jar extraction directories added in `extendJavaSourceModelWithSourcesJar` are never subject to exclusion — so if a third-party artifact publishes a sources jar that JavaParser chokes on, the downstream plugin build breaks with no escape hatch, in exchange for `@since`/`@deprecated`/description text.

Warn-and-skip per failed file, indexing whatever parsed, seems the safer default — at minimum for sources-jar roots. Failing hard on the project's own sources is defensible.

## Smaller notes

- `JarTypeSolver` never releases its jar handles: each one holds a javassist `ClassPool` whose `JarClassPath` keeps the `JarFile` open with no close API, and `JavaSourceModel.close()` only closes the `URLClassLoader`. Handles are released at GC. That is an upstream limitation, not something this PR can fully fix, but it does strengthen the case for fixing finding 1.
- `LinkUtils` (lines 54, 78) catches only `IllegalArgumentException`, but `getAllAncestors()`/`getAllFields()` throw `UnsolvedSymbolException` when an ancestor of a referenced type is not on the plugin classpath. The outer `catch (Throwable)` in `JavadocInlineTagsToXhtmlConverter.java:74` keeps the build alive, but output degrades from a rendered link to the original tag plus a stack trace warning, where QDox rendered these via phantom classes. Mapping `UnsolvedSymbolException` onto the unresolved-link path would be tidier.
- Performance: every `.java` file in every root is parsed eagerly and in full, all compilation units are retained for the duration of `execute()`, and symbol solving re-parses files through `JavaParserTypeSolver`'s separate cache. Much heavier than QDox on large trees and sources jars. Not a blocker; `tryToParseParallelized` and/or lazier parsing would be a reasonable follow-up.
- Behaviour deltas worth a release note: a null request encoding now defaults to UTF-8 (line 258) where QDox used the platform default; `{@value}` on binary-only constants no longer resolves, and initialisers are pretty-printed rather than reproduced verbatim; inherited field references are now attributed to the declaring supertype (this one matches real javadoc, so an improvement).
- Licensing is fine: JavaParser is dual LGPL-3.0/Apache-2.0, Category A under the Apache election, and nothing is bundled, so LICENSE/NOTICE are unaffected. Footprint does grow though — `javaparser-symbol-solver-core` pulls in guava, javassist and checker-qual (~5 MB against QDox's ~350 KB). QDox itself is cleanly gone from the reactor.

## Not your bug: the `gh-944` IT now fails on Windows

Changing the IT's `excludedScanDirectory` to `${project.basedir}/src/main/java` and adding a `verify.groovy` that actually asserts the exclusion took effect has exposed a **pre-existing bug in `DefaultPluginToolsRequest.isExcludedScanDirectory`** (`DefaultPluginToolsRequest.java:256-265`), which this PR does not touch:

```java
sourceFs.getPathMatcher("glob:" + excludedScanDirectory).matches(sourcePath)
```

In `glob:` syntax the backslash is an escape character and is consumed — `sun.nio.fs.Globs.toRegexPattern` applies this on Windows too. On Windows `${project.basedir}` interpolates to `D:\a\...`, so the pattern degenerates to `D:a...` and can never match a real path. The exclusion silently does nothing, `FirstMojo`'s Javadoc is scanned anyway, and the new assertion fails. The old IT used `${project.build.directory}/generated-sources/annotations` and never asserted the effect, which is why this stayed hidden.

So `excludedScanDirectories` has most likely never worked on Windows for any interpolated path. That deserves its own issue and fix rather than being papered over here — happy to take it. In the meantime the IT will keep this PR red on Windows.

wilx:
I have addressed some of the remaining review issues. I left out the _5. Member-only reference lookup no longer searches nested classes_ issue as it says the old behaviour is not conforming.

slachiewicz (approved):

slachiewicz:
Thank you!

PR #1189, opened 2026-09-08
Implement bootstrap module for plugin descriptor generation

fridrich:
Following this checklist to help us incorporate your
contribution quickly and easily:

- [x] Your pull request should address just one issue, without pulling in other changes.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x] Each commit in the pull request should have a meaningful subject line and body.
  Note that commits might be squashed by a maintainer on merge.
- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
  This may not always be possible but is a best-practice.
- [x] Run `mvn verify` to make sure basic checks pass.
  A more thorough check will be performed on your pull request automatically.
- [x] You have run the integration tests successfully (`mvn -Prun-its verify`).

If your pull request is about ~20 lines of code you don't need to sign an
[Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
please ask on the developers list.

To make clear that you license your contribution under
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

- [x] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
- [ ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

### DIFF
--- a/maven-plugin-plugin-bootstrap/pom.xml
+++ b/maven-plugin-plugin-bootstrap/pom.xml
@@ -1,0 +1,202 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one
+  ~ or more contributor license agreements.  See the NOTICE file
+  ~ distributed with this work for additional information
+  ~ regarding copyright ownership.  The ASF licenses this file
+  ~ to you under the Apache License, Version 2.0 (the
+  ~ "License"); you may not use this file except in compliance
+  ~ with the License.  You may obtain a copy of the License at
+  ~
+  ~   http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing,
+  ~ software distributed under the License is distributed on an
+  ~ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  ~ KIND, either express or implied.  See the License for the
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.apache.maven.plugin-tools</groupId>
+    <artifactId>maven-plugin-tools</artifactId>
+    <version>4.0.0-beta-4-SNAPSHOT</version>
+  </parent>
+
+  <groupId>org.apache.maven.plugins</groupId>
+  <artifactId>maven-plugin-plugin-bootstrap</artifactId>
+  <packaging>jar</packaging>
+
+  <name>Maven Plugin Plugin Bootstrap</name>
+  <description>Bootstrap module for generating Maven plugin descriptors outside Sisu/Plexus container.</description>
+
+  <properties>
+    <maven.install.skip>true</maven.install.skip>
+    <maven.deploy.skip>true</maven.deploy.skip>
+    <gpg.skip>true</gpg.skip>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-core</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven.plugin-tools</groupId>
+      <artifactId>maven-plugin-tools-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven.plugin-tools</groupId>
+      <artifactId>maven-plugin-tools-generators</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.ow2.asm</groupId>
+      <artifactId>asm</artifactId>
+    </dependency>
+
+    <!-- runtime extractors used by default by plugin-tools -->
+    <dependency>
+      <groupId>org.apache.maven.plugin-tools</groupId>
+      <artifactId>maven-plugin-tools-annotations</artifactId>
+    </dependency>
+    <!-- for source code annotations of the mojos -->
+    <dependency>
+      <groupId>org.apache.maven.plugin-tools</groupId>
+      <artifactId>maven-plugin-annotations</artifactId>
+    </dependency>
+
+    <!-- Maven -->
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-plugin-api</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-model</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-settings</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-repository-metadata</artifactId>
+      <version>${maven3Version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-artifact</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven.resolver</groupId>
+      <artifactId>maven-resolver-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven.resolver</groupId>
+      <artifactId>maven-resolver-util</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>javax.inject</groupId>
+      <artifactId>javax.inject</artifactId>
+      <version>1</version>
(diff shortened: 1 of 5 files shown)


fridrich:
This adds a profile called bootstrap. This profile is not run by default. When it is activated, the module `maven-plugin-plugin-bootstrap` is built. It is composed of all the sources of `maven-plugin-plugin` module and two additional files: the `org/apache/maven/plugin/plugin/StandaloneDescriptorGenerator.java` and its unit test.
This is basically a class that uses already built modules to generate a minimal plugin descriptor plugin.xml, which allows this module to be run as maven plugin. Its packaging is `jar` in order to avoid the lifecycles that come with the `maven-plugin` packaging. Once this is built, the module `maven-plugin-plugin` will use it for the generation of its HelpMojo.java and its descriptors.
This uses the fact that maven is able to use its own reactor modules in the build of further modules. The mojos run from this bootstrap plugin will be the same mojos that will be called and the generated xml files and the HelpMojo.java will be identical to ones that would be generated by a non-bootstrap plugin. The `maven-plugin-report-plugin` will then use the non-bootstrap plugin for its own lifecycle.
The `maven-plugin-plugin-bootstrap` artifact is guarded against installation to the local maven repository and against publishing or deployment. It is purely useful for the bootstrapping. And bootstrapping it does indeed. I checked, when using this bootstrap profile with clean local maven repository, none of the `maven-plugin-tools` artifacts were installed. All were compiled without needing any of their older version.
The StandaloneDescriptorGenerator class contains only code that is needed for generating a plugin descriptor that allows the bootstrap plugin to be ran. With a notable exception that I left in methods that allow generation of the V4 index and the mojo factory classes, if once it is needed so that nobody needs to reinvent a wheel that I already invented.

fridrich:
I run the integration tests in both bootstrap and non-bootstrap modes and I compared the content of the maven-plugin-plugin in both cases and the difference was basically the "Generated by" version. All the javadoc @see and similar were taken in account correctly and the resulting maven-plugin-plugin is a full-fledged plugin.

slawekjaranowski:
What problem do you try to resolve?

What is a use case for it?

fridrich:
> What problem do you try to resolve?
> 
> What is a use case for it?

As a distro packager, I come to this kind of situation when a package has circular dependency on itself, like maven-plugin-plugin module has. In a situation when one can download the whole internet, those kind of things are solvable by using an artifact with lower version. But this self-reference is a complete nightmare for distro packagers. It was a spine in my heel for quite a long time. So, once I decided to tackle it and came with this solution that has several arguments for itself:
- It's fully opt-in: -Pbootstrap profile, extra module only added to the reactor under that profile. It has zero impact on the default build, release process, or DescriptorGeneratorMojo/HelpGeneratorMojo themselves.
- It solves a genuine problem. Any distro/packaging system doing from-scratch bootstrap of the Maven plugin ecosystem hits the exact same maven-plugin-plugin self-reference wall. This isn't openSUSE specific, so Fedora or other RPM-based distros doing similar work would likely find it useful too.
- The mechanism is clever but not exotic or hacky. Reactor-plugin-resolution across sibling modules is a real Maven feature, not a hack against Maven's internals.

I had also other version of the standalone generator where I was able to reproduce almost 1:1 identical plugin.xml and completely identical HelpMojo.java as well as the V4 factories and indices. But that would be more intrusive and I preferred to trim it down to this, since inside xmvn context, I am able to build a random plugin by a generated ant build and generate the HelpMojo.java by running plugin:helpmojo outside of maven lifecycle and after the build generate the descriptors by running plugin:descriptor outside of the lifecycle too.

I know that this is a long shot with trying to upstream it, but I tried to make it as minimal as possible.

slachiewicz:
Read through the module. The problem it solves (a distro cannot build maven-plugin-plugin without a prior maven-plugin-plugin) is real, but this shape is hard to keep alive: it injects into four private fields of `JavaAnnotationsMojoDescriptorExtractor` by reflection, so any rename breaks it at runtime and only under `-Pbootstrap`, which CI does not run; it re-implements POM parent walking, property interpolation and dependency resolution with placeholder `dummy.jar` files; and it compiles all of `maven-plugin-plugin` a second time through `compileSourceRoots`. Before this can be considered I would want: a CI job that runs the profile, package-private constructors instead of reflection, and ideally a much smaller approach that boots the real mojo through a minimal Sisu container instead of duplicating its wiring. Converting to draft until then.

fridrich:
> Before this can be considered I would want:
> a CI job that runs the profile

added

> package-private constructors instead of reflection, and ideally a much smaller approach that boots the real mojo through a minimal Sisu container instead of duplicating its wiring.

originally, my goal was not to add any code to maven-plugin-plugin to keep it completely pristine in order to leave the non-bootstrap path as untouched as possible. Your review actually helped me to come with a solution that is IMHO even more elegant even though it adds a main method to the org.apache.maven.plugin.plugin.DescriptorGeneratorMojo class. If this is good for you, I find it superior solution.

fridrich:
> originally, my goal was not to add any code to maven-plugin-plugin to keep it completely pristine in order to leave the non-bootstrap path as untouched as possible. Your review actually helped me to come with a solution that is IMHO even more elegant even though it adds a main method to the org.apache.maven.plugin.plugin.DescriptorGeneratorMojo class. If this is good for you, I find it superior solution.

Actually, no, it will not work like this. Have to change something more.

fridrich:
> Read through the module. The problem it solves (a distro cannot build maven-plugin-plugin without a prior maven-plugin-plugin) is real, but this shape is hard to keep alive: it injects into four private fields of `JavaAnnotationsMojoDescriptorExtractor` by reflection, so any rename breaks it at runtime and only under `-Pbootstrap`, which CI does not run; it re-implements POM parent walking, property interpolation and dependency resolution with placeholder `dummy.jar` files; and it compiles all of `maven-plugin-plugin` a second time through `compileSourceRoots`. Before this can be considered I would want: a CI job that runs the profile, package-private constructors instead of reflection, and ideally a much smaller approach that boots the real mojo through a minimal Sisu container instead of duplicating its wiring. Converting to draft until then.

So, this fix addresses your comments:

1) *CI job for -Pbootstrap*: I added a bootstrap job in maven-verify.yml. It runs mvn -B -Pbootstrap verify on every push/PR, so now we test this profile.

2) *Reflection into private fields*: Removed completely, not just replaced with a constructor. `StandaloneDescriptorGenerator` does not create `JavaAnnotationsMojoDescriptorExtractor` manually anymore. Now it starts a small `DefaultPlexusContainer` and calls `container.lookup(MojoScanner.class)`. This way Sisu connects the extractor like a normal Maven build (real @Inject fields, no reflection, no manual setup).

3) *Small Sisu container instead of fake code*: The same container lookup also replaces the custom fake `RepositorySystem` (which was a dynamic proxy that only worked for `newLocalRepositoryManager`) and the custom `ArchiverManager` implementation. Both now come from the real container.

What I did not change on purpose: the local, offline parent POM reader (property interpolation and fake-jar placeholders) that creates the descriptor for `maven-plugin-plugin-bootstrap`. I tried to change it to use real `ProjectBuilder`, but that tries to download and resolve the whole parent POM chain (like junit-bom, which we don't need here). It fails when local repo is empty because our manual container doesn't have the HTTP connector configured like Maven CLI. So I reverted this change. This module's descriptor only needs to work enough for the plugin to run; nobody else uses it. The real descriptor for `maven-plugin-plugin` is generated later by normal Maven process after `maven-plugin-plugin-bootstrap` is ready. I checked it, and it has the correct description, correct `requiredJavaVersion`/`requiredMavenVersion`, and correct dependencies.

The extra `compileSourceRoot` for `maven-plugin-plugin` sources (compiling them twice) is still there. This is needed because the bootstrap module makes a copy of the mojo classes to avoid circular dependency, and I couldn't find a smaller way to fix it.

I tested with empty local repo (-Dmaven.repo.local): full build passes, and it doesn't download any artifacts from this project (`maven-plugin-tools-*`, `maven-plugin-plugin*`).

fridrich:
For the bootstrap descriptor, one does really need only a minimal repository system to satisfy sisu `@Inject` (JavaAnnotationsMojoDescriptorExtractor.java:150 declares `@Inject private RepositorySystem repositorySystem;`) and it is enough for it to throw and not to do anything. The ArtifactResolutionException is caught and all works fine after. This avoids full aether/resolver runtime and for bootstrap descriptor things like `@since` or `@deprecated` are not necessary.

fridrich:
The stray newline, I discovered when comparing the plugin descriptors generated by 3.15 in non-bootstrap build and with own 4.0.x in the bootstrap one. The reason was that jsoup inserts new line after HTML comments in javadoc conversion which ends up as a extra space when converted to plain text. The last fix makes the plugin descriptors identical besides the "Generated by ..." version which is normal.

fridrich:
This commit somehow removes the necessity of hand-crafted walking of pom hierarchy.

fridrich:
The last commit deduplicates the v4 DI-index generation. I made it separate last commit because it touches maven-plugin-plugin code and if you are unconfortable with it, it can be dropped readily, since edits by maintainers are allowed in this PR.

fridrich:
Here I leave it and will not try to improve more, since I am able to improve it to worse. Waiting for your feedback

fridrich:
I updated the stray newline commit with changes to the integration tests, basically reverting 24420ee6260c6ce95e671d0f74a2f2f63b317d4a. Now the integration tests pass when -Pbootstrap is specified and also in the normal situation.

PR #1209, opened 2026-09-20
switch from 4.0.0-beta-4 to 3.17.0

hboutemy:
see https://github.com/apache/maven-sources/issues/57#issuecomment-5750294987

let's maintain only `master` that produces a Maven 3 plugin being able to generate both Maven 3 plugin descriptors and Maven 4 plugin descriptors

this acts also dropping `maven-script` and old-Maven 2-style javadoc annotations


hboutemy:
need to check/clarify Maven 3 plugin.xml generation = https://maven.apache.org/ref/3.9.16/maven-plugin-api/plugin.html
vs Maven 4 one = https://maven.apache.org/ref/4.0.0-rc-6/api/maven-api-plugin/plugin.html
see also https://github.com/apache/maven/pull/13253

key difference is not only the new fields in Maven 4 plugin descriptor, but renamings:
- `requiresDependencyResolution` to `dependencyResolution`
- `require*` to new names
see https://github.com/apache/maven/blob/master/api/maven-api-plugin/src/main/mdo/plugin.mdo#L211

PR #1212, opened 2026-09-22
Add Apache prefix to ATR project name

sbp:
Add a missing "Apache " prefix to `project.metadata.name` in `.asf.yaml`. ATR requires project display names to start with "Apache ". OF, therefore not reviewed beyond the trivial change.


hboutemy:
thanks

